PR #103, opened 2015-05-03
upgpkg: pycharm-professional 4.0.6

alviezhang:
Update pycharm-professional to version 4.0.6


lilydjwg:
Changing the PKGBUILD alone won't update the package (unless it has [lilac](https://github.com/archlinuxcn/lilac) enabled). So you'll have to wait until the maintainer packages the new one.

PR #200, opened 2016-02-15
New package: rak - Grep replacement in pure Ruby

farkasity:


lilydjwg:
You need to provide a `lilac.py` when you open PRs, or there won't be any package in the repository even merged.

However I've seen your application mail. I'll process it soon.

farkasity:
I have added missing lilac.py, but I thought it maybe inconvenient, maybe we could trying to simplify it by reflect lilac.

lilydjwg:
Go ahead and send PRs to [lilac](https://github.com/archlinuxcn/lilac) then :-)

PS: using `aur_post_build` might cause problems; there are (or were) other gem packages in the repo and you can take a look at them.

farkasity:
OK. I will optimize lilac.py latterly.

Issue #474, opened 2017-08-17
Package request: CopyQ


chih7:
CopyQ is advanced clipboard manager with editing and scripting features.
based on qt5, I think it is better than Klipper.

[GitHub](https://github.com/hluk/CopyQ)
[AUR](https://aur.archlinux.org/packages/copyq/)

chih7:
I submitted a pull request in https://github.com/archlinuxcn/repo/pull/475

PR #475, opened 2017-08-17
Add copyq

chih7:
fix https://github.com/archlinuxcn/repo/issues/474


lilydjwg:
LGTM, let's see how it works out.

PR #503, opened 2017-09-17
Update PKGBUILD

milkice233:
No need to check the changing ToS file


lilydjwg:
This isn't going to work. Changing the PKGBUILD here doesn't automatically update the built package.

PR #533, opened 2017-10-12
add otter-browser

m4sk1n:


lilydjwg:
You can't pull request for packages without lilac support, because no one will be actually building and uploading the packages.

m4sk1n:
@lilydjwg and now?

lilydjwg:
Let lilac try it :-)

PR #550, opened 2017-10-31
pre-commit check index tree instead of working tree

farseerfc:
This change on pre-commit hook tries to check the index tree (files already added by `git add`) instead of the working tree. So that it will not see the ignored/removed files (as they not being in the index tree), and will detect mistakes like forgetting to add modified nvchecker.ini.

The down side: `git commit` will take more than 5 secs on my SSD machine, because all the `popen("git", ...)` costs. I already tried `lru_cache` to save some time. It was more than 10s before using lru_cache.

Should we use this new hook or stick to the old one (which is more quicker) ?

### DIFF
--- a/pre-commit
+++ b/pre-commit
@@ -8,17 +8,20 @@ import sys
 import re
 from subprocess import Popen, PIPE
 from unittest import TestCase, TestSuite, TestLoader, TextTestRunner, skip
-from io import StringIO
+from io import TextIOWrapper, StringIO
 from configparser import ConfigParser
-from functools import lru_cache
 
+DIRS, FILES = None, None
 GIT_BRANCH = 'HEAD'
+
 
 def readutf8(st):
     return st.read().decode('utf-8').strip()
 
+
 def reads(st):
     return readutf8(st).strip()
+
 
 def git_write_tree_index():
     global GIT_BRANCH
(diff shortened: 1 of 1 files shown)


lilydjwg:
use pygit2?

lilydjwg:
另外用 `git ls-tree -r` 也不错的。

farseerfc:
不太想加額外依賴，我試試改成只用一次 `git ls-tree -tr`

pre-commit
```diff
@@ -6,120 +6,165 @@
...
+    cmd = ['git', 'write-tree']
+    with Popen(cmd, stdout=PIPE) as p:
+        GIT_BRANCH = reads(p.stdout)
+
+git_write_tree_index()
+
+def git_open(filename):
+    cmd = ['git', 'cat-file', 'blob', f'{GIT_BRANCH}:{filename}']
+    with Popen(cmd, stdout=PIPE) as p:
+        return StringIO(readutf8(p.stdout))
```
lilydjwg: Why not `return io.TextIOWrapper(p.stdout, encoding='utf-8')`?

pre-commit
```diff
@@ -6,120 +6,165 @@
...
+
+
+@lru_cache(maxsize=1 << 16, typed=True)
+def git_listdir(path):
+    cmd = ['git', 'ls-tree', '--name-only',
+           (f'{GIT_BRANCH}:{path}' if path != '.' else f'{GIT_BRANCH}')]
```
lilydjwg: 后边不需要 f'' 了。

lilydjwg:
没动静了？

swordfeng:
先读修改的文件列表，构建一个匹配自动机，如果要读的文件在列表中再去用git命令读取，否则直接读文件系统的文件，如何？

pre-commit
farseerfc: Because `p.stdout` will have been closed when we read from this later on

farseerfc:
用 `git ls-tree` 加快了 isdir 和 isfile 的速度，現在只做一次 `ls-tree` 然後把結果存在了數組裏。
加速之後運行時間在我這邊大概 0.33s ，相比於 master 那個 0.26s 幾乎沒有性能損失了。

pre-commit
```diff
@@ -6,120 +6,185 @@
...
+DIRS, FILES = None, None
+GIT_BRANCH = 'HEAD'
+
+
+def readutf8(st):
+    return st.read().decode('utf-8').strip()
+
+
+def reads(st):
+    return readutf8(st).strip()
```
lilydjwg: 这里不是已经 strip() 过的吗？

farseerfc: git ls-tree -z 的話行結尾是 nul ，好像不會被 strip 掉。不考慮特殊文件名的話可以不用 -z ，目前好像repo裡面文件名只有空格比較特殊。

farseerfc:
刪掉了不需要的 strip 和 reads

PR #613, opened 2017-12-18
Add intellij-idea-ultimate-edition

Sherlock-Holo:


Sherlock-Holo:
`intellij-idea-community-edition` is too old and then I want to use the new version so I want to add it to CN repo

PR #616, opened 2017-12-19
webkitgtk: update from AUR; fixed building with enchant 2.x

yan12125:
This is basically the same as https://aur.archlinux.org/cgit/aur.git/commit/?h=webkitgtk&id=42cc417a1edb3f36eef984a12f9f059f242ea19b except the ```epoch``` part.

cc @colinkeenan

### DIFF
--- a/webkitgtk/PKGBUILD
+++ b/webkitgtk/PKGBUILD
@@ -4,7 +4,7 @@
 pkgbase=webkitgtk
 pkgname=(webkitgtk webkitgtk2)
 pkgver=2.4.11
-pkgrel=6
+pkgrel=7
 epoch=3
 pkgdesc="Legacy Web content engine"
 arch=(i686 x86_64)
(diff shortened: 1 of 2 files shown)


colinkeenan:
I have been building it for several hours now, but had forgotten that I added epoch=3 to it in the past. I guess I'll start over.

lilydjwg:
@colinkeenan you can use the build machine, which will be much faster.

colinkeenan:
I don't know how.

lilydjwg:
@colinkeenan You still have your private key, right? `ssh colinkeenan@build.archlinuxcn.org`, then put your PKGBUILD etc somewhere, run `extra-x86_64-build` instead of `makepkg` to build.

colinkeenan:
I will look at that next time.Now going to sleep while it's building. It will be done when I get up.

colinkeenan:
I uploaded webkitgtk and webkitgtk2 twice. The first time, I accidentally uploaded a bad signature file that was made against both of them.

Issue #671, opened 2018-03-05
Package Request: boostnote


chih7:
### 问题类型 / Type of issues

* 软件打包请求 / package request

### 受影响的软件包 / Affect packages

* boostnote

----

* 简述 / Brief Description

Open source note-taking app for programmers

* 主要语言和工具链 / Main Languages and Toolchains

electron nodejs

* 上游链接 / Upstream Link

https://github.com/BoostIO/Boostnote
https://aur.archlinux.org/packages/boostnote/

* 为何需要这个软件包 / Why Need This Package

Useful note software on the Linux desktop platform

chih7:
I try to create a pull request https://github.com/archlinuxcn/repo/pull/672 to solve this.

PR #672, opened 2018-03-05
addpkg: boostnote

chih7:
I try to create a pull request to solve https://github.com/archlinuxcn/repo/issues/671.


nvchecker.ini
```diff
@@ -2495,6 +2495,9 @@ aur =
 
 [fcitx5-anthy-git]
 aur =
+
```
lilydjwg: Add your name and email address here like above. This will be used by lilac in the future to mail error reports.

Issue #805, opened 2018-07-16
Out-of-date: google-earth-pro


gojpdchx:
### 问题类型 / Type of issues

* 过期软件包 / out-of-date packages

### 受影响的软件包 / Affected packages

* google-earth-pro

----
The AUR package is now at version 7.3.2.5487, but [archlinuxcn] has the old version 7.3.0.3832.

lilacbot:
build log for auto building out-of-date packages:
```
[2018-07-12 01:41:49] google-earth-pro 7.3.2.5487-1 failed after 3s
```

PR #809, opened 2018-07-16
Update google-earth-pro to 7.3.2

heavysink:
By using curl query, lilac will automatically update it by checking the new version from Google.


heavysink:
@gojpdchx #805

PR #897, opened 2018-10-28
update systemtap to 4.0

ideal:
Need to change too much so lilacbot autoupdate failed.

### DIFF
--- a/systemtap/PKGBUILD
+++ b/systemtap/PKGBUILD
@@ -13,9 +13,12 @@ makedepends=('python2-setuptools' 'xmlto' 'cpio')
 optdepends=('sqlite3: for storing results in a database')
 source=("${pkgname}-${pkgver}.tar.gz::https://sourceware.org/systemtap/ftp/releases/${pkgname}-${pkgver}.tar.gz"
         "${pkgname}-${pkgver}.tar.gz.asc::https://sourceware.org/systemtap/ftp/releases/${pkgname}-${pkgver}.tar.gz.asc"
+        "systemtap.sysusers"
         )
 sha512sums=('66b5ba1902ed974101429dce9fee849273d3dc64e862c057177d158c5f284cc93299819ee8ee49d4bd3b7ff47a3512b1c68b74cad374147ce67fa406c7518c79'
-            'SKIP')
+            'SKIP'
+            '0b54000f0bb6ed9eb4f43dbdc2fd66fcc178fc24688610fcb7b2ace5195bf96a956e24f2c281545db1a5a7c98aa951af63e8bd00de1f0347bae6f379f3671344'
+            )
 install='systemtap.install'
 validpgpkeys=('5D38116FA4D3A7CC77E378D37E83610126DCC2E8')
 
@@ -44,5 +47,8 @@ package() {
   make DESTDIR="${pkgdir}" install
   rm -rf "${pkgdir}/var"
   rm -rf "${pkgdir}/etc"
+
+  cd "${srcdir}"
+  install -Dm644 ${pkgname}.sysusers "${pkgdir}"/usr/lib/sysusers.d/${pkgname}.conf
 }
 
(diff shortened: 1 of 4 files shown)


systemtap/PKGBUILD
```diff
@@ -2,30 +2,30 @@
...
 
 build() {
   cd "${pkgname}-${pkgver}"
   sed -i -e "42,50d" stapbpf/Makefile.am
-  sed -i -e "840,848d" stapbpf/Makefile.in
-  sed -i -e "73,80d" staprun/Makefile.am
-  sed -i -e "1110,1117d" staprun/Makefile.in
+  sed -i -e "842,850d" stapbpf/Makefile.in
+  sed -i -e "77,84d" staprun/Makefile.am
+  sed -i -e "1115,1122d" staprun/Makefile.in
```
yan12125: Using line numbers is fragile and hard to maintain. Patches should be better solution.

By the way, [PKGBUILD for the same package on AUR](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=systemtap) does not have such lines. Are they necessary?

yan12125: IIUIC these lines will be executed as the `package()` function in PKGBUILD is run under fakeroot, so `id -u` return 0. In Arch Linux, it's encouraged to use `sysusers.d` instead of adding users and groups manually [1], so it's better to remove these lines from Makefiles.

[1] https://www.archlinux.org/todo/switch-to-systemd-sysusers/

ideal: Yes you are right, I tried with fakeroot `id -u` return 0. The removed lines in Makefile actually will take effect in `package()`. Sorry for my misunderstanding.

ideal: I checked [this][1], is below a better way ?

1. do not remove lines in Makefiles, thus mode bits and group of `staprun` `stapbpf` will be already ok in generated package
2. add a systemtap.sysusers
3. remove group adding, chgrp, chmod in systemtap.install

Is this ok? not sure about that.

[1]: https://www.archlinux.org/todo/switch-to-systemd-sysusers/

yan12125: Oops I didn't notice there are `chown` commands. In this it might be better to keep the current approach and not moving to sysusers.

ideal: I've already changed. Do i need to close this PR and recreate another?

yan12125: You can just continue your work in this PR.

Let's wait for lijinpei before further decisions.

Xuanwo:
@yan12125 Do you have time to review this pr again ?

yan12125:
I think this PR will be ready if line number based deletion are replaced with applying patch files.

ideal:
@yan12125 OK.

ideal:
The sed operations are replaced now. Please review it. Thanks.

systemtap/systemtap.install
```diff
@@ -1,8 +1,4 @@
 post_install() {
-  getent group stapusr >/dev/null || groupadd -g 156 -r stapusr 2>/dev/null || groupadd -r stapusr
-  getent group stapsys >/dev/null || groupadd -g 157 -r stapsys 2>/dev/null || groupadd -r stapsys
-  getent group stapdev >/dev/null || groupadd -g 158 -r stapdev 2>/dev/null || groupadd -r stapdev
-
   getent group stapusr >/dev/null && chgrp stapusr /usr/bin/stapbpf
```
yan12125: `chgrp stapusr` should be `chgrp 156` as the `post_install()` function is run before the `.sysusers` file is processed.

yan12125: Furthermore, chgrp commands are better to be in `package()` of PKGBUILD.

yan12125: If `chgrp` commands are in `post_install()`, `pacman -Qkk systemtap` will report GID mismatch for these files. That's a minor issue, anyway.

ideal: OK, moved.

yan12125:
Merged. Thank you very much for the efforts and patience on the update!

PR #910, opened 2018-11-07
add i3blocks-gaps-git

masakichi:


nvchecker.ini
```diff
@@ -2900,6 +2900,9 @@ github = zsh-users/zsh-completions
 [zsh-syntax-highlighting-git]
 github = zsh-users/zsh-syntax-highlighting
```
a-wing: Please. Add your name and email if you use lilac.py

TobiichiAmane: Is it suggested to use `lilac.yaml` by default? Using `nvchecker.ini` may cause problems when switching to `lilac.yaml` later.

Issue #918, opened 2018-11-14
virtualbox-ext-oracle out of date


evansan:
### 问题类型 / Type of issues

* 过期软件包 / out-of-date packages

### 受影响的软件包 / Affected packages

* virtualbox-ext-oracle

lilydjwg:
The issuebot has a bug....(actually I've filed [a Python bug](https://bugs.python.org/issue35246).)

Issue #708, opened 2018-04-19
virtualbox-ext-oracle out of date


yan12125:
### 问题类型 / Type of issues

* 过期软件包 / out-of-date packages

### 受影响的软件包 / Affect packages

* virtualbox-ext-oracle

----

In nvchecker.ini, virtualbox-ext-oracle uses `pacman = virtualbox`.

https://github.com/archlinuxcn/repo/blob/0619edc6a67616aaf7bc2cf601c3d8de467f8598/nvchecker.ini#L325-L326

As a result, people have to wait for updates of pacman databases on the build machine to get updated virtualbox-ext-oracle.

Here are two options:
* Switch to `archpkg = virtualbox`
* Switch to `aur = `. The AUR package virtualbox-ext-oracle is now maintained by eworm, the TU who also maintains community/virtualbox. In recent versions, virtualbox-ext-oracle were all updated soon after community/virtualbox is bumped. For example, virtualbox 5.2.10-1 is signed on 2018-04-17 21:20 UTC and virtualbox-ext-oracle 5.2.10-1 is updated on 2018-04-17 21:34:42 UTC.

lilacbot:
build log for auto building out-of-date packages:
```
[2018-02-28 18:10:16] virtualbox-ext-oracle 5.2.8-1 successful
```

yan12125:
Well, 下次virtualbox-ext-oracle更新時應該還是會有類似的問題(這次archlinuxcn比aur晚兩天)，下次遇到再來處理好了

PR #919, opened 2018-11-14
virtualbox-ext-oracle: make it follow AUR

yan12125:
This package was tracking local pacman database. If local databases are
not up-to-date, this package will not be updated either. This is the
case on build.archlinuxcn.org as of now.

Closes: https://github.com/archlinuxcn/repo/issues/918

### DIFF
--- a/nvchecker.ini
+++ b/nvchecker.ini
@@ -310,7 +310,7 @@ pypi = uniout
 github = dubiousjim/dcron
 
 [virtualbox-ext-oracle]
-pacman = virtualbox
+aur =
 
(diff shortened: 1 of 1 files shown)


yan12125:
> But why not use `archpkg`?

As I mentioned in https://github.com/archlinuxcn/repo/issues/708#issue-315835776, aur/virtualbox-ext-oracle is well maintained. For this package, following AUR should lower maintanence burden than following the corresponding Arch package.

> Also could you use lilac.yaml?

I can do that later.

PR #925, opened 2018-11-23
Switch sasasu's package to lilac.yaml

Sasasu:


lilydjwg:
Don't delete lilac.py. You can use an empty file if you have nothing to write in it.

PR #935, opened 2018-11-28
migrate i3blocks-gaps-git to lilac.yaml

masakichi:


tobiichiamane:
It need an empty lilac.py. Because lilac will generate a part of lilac.py according to lilac.yaml, and merge into the original py file.

masakichi:
updated: keep the lilac.py empty.

PR #979, opened 2018-12-30
Feature/static dep check

VOID001:
Check static dependency errors as soon as possible in order to notify the developer basic configuration error in lilac.py and lilac.yaml. Avoid waiting for 8 hours until lilac start the job and complain the package fail to meet the depdency requirement.

https://travis-ci.org/archlinuxcn/repo/builds/473455675

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,8 @@
 !.travis.yml
 !.git*
 !.github/**
+!.lilaclib.py
+lilaclib.py
 pkg/
 src/
 *.xz
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,4 @@ linux.install.pkg
 *.bin
 *.exe
 *.ttf
+__pycache__/
--- a/.lilaclib.py
+++ b/.lilaclib.py
@@ -1,8 +1,0 @@
-def aur_pre_build(): pass
-def pre_build(): pass
-def pypi_pre_build(): pass
-def vcs_update(): pass
-def aur_post_build(): pass
-def post_build(): pass
-def post_build_always(): pass
-def pypi_post_build(): pass
--- a/.lilaclib.py
+++ b/.lilaclib.py
@@ -1,0 +1,8 @@
+def aur_pre_build(): pass
+def pre_build(): pass
+def pypi_pre_build(): pass
+def vcs_update(): pass
+def aur_post_build(): pass
+def post_build(): pass
+def post_build_always(): pass
+def pypi_post_build(): pass
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,6 @@ install:
 
 script:
     - python3 --version
-    - mv .lilaclib.py lilaclib.py
     - export PYTHONPATH=$PYTHONPATH:$(pwd)
     - ./pre-commit --all # checking all PKGBUILD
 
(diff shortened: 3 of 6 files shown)


VOID001:
Developers should move `.lilaclib.py` to `lilaclib.py`  in the repo root to use the `pre-commit` hook

yan12125:
> Developers should move `.lilaclib.py` to `lilaclib.py` in the repo root to use the pre-commit hook

A wild guess - adding some stuff in `sys.modules` might remove the necessity of this step.

yan12125:
> Yes, you can fake such a module.

I created it in 77660f3e990fad5f477ddaf9614556fd8fa25aac. I omit pre_build, post_build and post_build_always as they are provided by individual lilac.py files instead of lilaclib.

yan12125:
I attempted to rebase this branch onto the current master, but the `pre-commit` test crashes with segmentation fault. Any ideas?

lilydjwg:
You may want to [enable debug mode](https://docs.travis-ci.com/user/running-build-in-debug-mode/) and figure out what causes the segfault. Also you may want to use a newer Python.

yan12125:
Thanks for the info. Following the API call on that page always give me 403 forbidden. On the other hand, I can reproduce the crash locally on a 18.04 container. The last few frames in the stack trace are:
```
(gdb) bt
#0  0x0000000000000000 in ?? ()
#1  0x00007f2ca7bdc4a2 in ?? () from /usr/lib/x86_64-linux-gnu/libalpm.so.11
#2  0x00007f2ca7bdd007 in alpm_find_satisfier () from /usr/lib/x86_64-linux-gnu/libalpm.so.11
#3  0x00007f2ca7e03262 in pyalpm_find_satisfier (self=<optimized out>, args=<optimized out>) at src/pyalpm.c:55
```
And instructions around frame `#1` are:
```
(gdb) disassemble 0x00007f2ca7bdc49d,+10
Dump of assembler code from 0x7f2ca7bdc49d to 0x7f2ca7bdc4a7:
   0x00007f2ca7bdc49d:  callq  0x7f2ca7bcf220 <alpm_pkg_get_provides@plt>
=> 0x00007f2ca7bdc4a2:  mov    %rbx,%rdi
   0x00007f2ca7bdc4a5:  mov    %rax,%rsi
```
The PLT entry points to NULL. Sounds a broken runtime linker or corrupted memory.

lilydjwg:
> Following the API call on that page always give me 403 forbidden.

You need to mail their support team to enable that as the page says (and it is Saturday today...). But you've reproduced it locally so it's not needed :-)

yan12125:
> You need to mail their support team to enable that as the page says

Oops, didn't notice that.

No matter that crash is fixed or not, I'm rather suspectible about using pyalpm/libalpm natively on Ubuntu. IMO that brings much higher maintenance costs for the `pre-commit` script.

lilydjwg:
Yes, there would be a lot of things to maintain. I prefer to use a container so we don't need to mess up with Ubuntu and deb packages.

SilverRainZ:
Any progress? or considering close it?

VOID001:
any updates on this pr？ if no then I will close it

yan12125:
It's up to the author to keep this PR open or not~

Anyway, this PR needs non-trivial changes after CI migrates from Travis to GitHub actions.

PR #994, opened 2019-01-06
tree-wide: fix incorrect vcs specs in update_on

yan12125:
Most likely they are victims of a nvchecker2yaml bug.


yan12125:
Thanks for looking into this! Yep the fixed version is equivalent to `vcs =`, which informs nvchecker to use the VCS URL from PKGBUILD.

PR #1089, opened 2019-03-19
python-torchvision: update lilac.yaml & lilac.py, also add quote to $var

hubutui:
now update on github release.


edward-p:
> looks good to me.

+1

PR #1096, opened 2019-03-23
jdownloader2: remove an unused file

yan12125:
Both JDownloader and jdownloader are in the repo, which confuses case-insensitive file systems like APFS. (I sync the archlinuxcn repo across all my devices, so this repo appears on both EXT4 and APFS.)


heavysink:
非常感谢！

PR #1113, opened 2019-04-05
add python-skorch

hubutui:
A scikit-learn compatible neural network library that wraps pytorch.


dctxmei:
This seems worth trying.

PR #1114, opened 2019-04-05
add python-augmentor

hubutui:
Image augmentation library in Python for machine learning


dctxmei:
The test was successful. By the way, your blog is not bad.

PR #1149, opened 2019-04-30
add 3dslicer-bin

hubutui:
A multi-platform, free open source software package for visualization and medical image computing


archlinuxcn/3dslicer-bin/lilac.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env python3
+
+from lilaclib import *
+import subprocess
+
+def pre_build():
+  update_pkgver_and_pkgrel(_G.newver.lstrip('v'))
+  cmd = "curl https://download.slicer.org/ |grep bitstream |cut -d '/' -f 3 |cut -d '\"' -f 1 |sed -n 3p"
```
edward-p: Try to use escape character `\'` for `'` here
`cmd = 'curl https://download.slicer.org/ |grep bitstream |cut -d \'/\' -f 3 |cut -d \'"\' -f 1 |sed -n 3p'`

archlinuxcn/3dslicer-bin/lilac.py
```diff
@@ -0,0 +1,23 @@
...
+  newstr = subprocess.run(cmd, stdout=subprocess.PIPE, shell=True).stdout.decode('utf-8')[0:-1]
+
+  cmd = "grep bitstream PKGBUILD|cut -f 5 -d '/'
+  oldstr = subprocess.run(cmd, stdout=subprocess.PIPE, shell=True).stdout.decode('utf-8')[0:-2]
+
+  if oldstr == newstr:
+    raise RuntimeError("URL is not updated yet.")
```
lilydjwg: You can use a `prepare` function to check whether a build should be done, and return a string when it's not ready (example in rust-nightly). You don't receive error reports for skipped builds.

Also, you can use Python to retrieve the URL and update PKGBUILD. There is a requests.Session object named `s` available to send requests. There is a `edit_file` function for you to update PKGBUILD easily.

archlinuxcn/3dslicer-bin/lilac.py
```diff
@@ -0,0 +1,23 @@
...
+  cmd = f"sed -i '/bitstream/s/{oldstr}/{newstr}/' PKGBUILD"
```
lilydjwg: This doesn't look safe. Please change it.

archlinuxcn/3dslicer-bin/lilac.py
```diff
@@ -5,15 +5,16 @@
...
+  newstr = newstr.split('\n')[2].replace('/bitstream/', '').replace('/bitstream/', '')
 
-  cmd = "grep bitstream PKGBUILD|cut -f 5 -d '/'
+  cmd = "grep bitstream PKGBUILD|cut -f 5 -d \'/\'"
   oldstr = subprocess.run(cmd, stdout=subprocess.PIPE, shell=True).stdout.decode('utf-8')[0:-2]
 
   if oldstr == newstr:
     raise RuntimeError("URL is not updated yet.")
-  cmd = f"sed -i '/bitstream/s/{oldstr}/{newstr}/' PKGBUILD"
+  cmd = f"sed -i \'/bitstream/s/{oldstr}/{newstr}/\' PKGBUILD"
```
lilydjwg: This doesn't solve the issue when `oldstr` or `newstr` contains unexpected content.

Please, avoid using `shell=True` when there is input from outside.

PR #1192, opened 2019-05-26
add riscv-none-embed

sequencer:
1. use official gnu upstream, rather than SiFive.
2. patch config.sub with embed in os.
3. disable libgloss default linking.

### DIFF
--- a/archlinuxcn/riscv-none-embed-binutils/0001-Add-support-for-SiFive-CLIC-CSRs.patch
+++ b/archlinuxcn/riscv-none-embed-binutils/0001-Add-support-for-SiFive-CLIC-CSRs.patch
@@ -1,0 +1,105 @@
+From 3613b170bb24afaa84f8064055fc84c6f07f92a4 Mon Sep 17 00:00:00 2001
+From: Jim Wilson <jimw@sifive.com>
+Date: Thu, 21 Feb 2019 19:21:23 -0800
+Subject: [PATCH 1/2] Add support for SiFive CLIC CSRs.
+
+---
+ gas/testsuite/gas/riscv/priv-reg.d |  4 ++++
+ gas/testsuite/gas/riscv/priv-reg.s |  6 ++++++
+ include/opcode/riscv-opc.h         | 19 +++++++++++++++++++
+ 3 files changed, 29 insertions(+)
+
+diff --git a/gas/testsuite/gas/riscv/priv-reg.d b/gas/testsuite/gas/riscv/priv-reg.d
+index 9ec5d97ed8..964c99412d 100644
+--- a/gas/testsuite/gas/riscv/priv-reg.d
++++ b/gas/testsuite/gas/riscv/priv-reg.d
+@@ -251,3 +251,7 @@ Disassembly of section .text:
+ [ 	]+3c4:[ 	]+3bd02573[ 	]+csrr[ 	]+a0,pmpaddr13
+ [ 	]+3c8:[ 	]+3be02573[ 	]+csrr[ 	]+a0,pmpaddr14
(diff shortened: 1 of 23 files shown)


sequencer:
Currently, riscv toolchain is using gcc with `libgloss` for both embed device and Linux, which is vended by SiFive, with the prefix `riscv64-unknown-elf`.
This cause too much trouble for embedded device usage, and [gnu-mcu-eclipse](https://github.com/gnu-mcu-eclipse/riscv-none-gcc) tried to fix the `libgloss` problem. This package takes the idea of gnu-mcu-eclipse to add `embed` to OS in `config.sub`, and remove the `libgloss` default linking.  
However, gnu-mcu-eclipse still stick on the [riscv-gcc](https://github.com/riscv/riscv-gcc/) branch, which didn't update for a long time. Since RISC-V has been upstreamed for since gcc7. This package patches the GNU upstream to provide the newest version of `riscv-none-embed`.

Sasasu:
please add `lilac.py` and `lilac.yaml` for each package.  The auto build system uses those files to monitoring a package.

lilac.py is a historical problem, just put an empty file.

lilac.yaml describe when can this package update or rebuild.

yaml schema file for lilac.yaml: https://github.com/archlinuxcn/repo/blob/master/lilac-yaml-schema.yaml 
manually update example: https://github.com/archlinuxcn/repo/search?l=YAML&q=manual 
track aur example: https://github.com/archlinuxcn/repo/search?l=YAML&q=aur_pre_build

ilg-ul:
> However, gnu-mcu-eclipse still stick on the riscv-gcc branch

this is not accurate, gnu-mcu-eclipse now follows the SiFive fork https://github.com/sifive/riscv-gcc.

> use official gnu upstream, rather than SiFive.

by doing this you miss the SiFive extensions (like interrupt attributes), and you cannot use your toolchain to build the SiFive Freedom SDK projects.

if you insist on doing this, I think you should explicitly mention this difference in the package metadata.

sequencer:
OK, I’ll collect a set of patch from SiFive gcc and maintain a patch to upstream.

ilg-ul:
I think this is a better approach.

Also keeping a consistent functionality with **GNU MCU Eclipse RISC-V Embedded GCC** (which will be soon renamed **xPack RISC-V Embedded GCC** and moved from **GNU MCU Eclipse** to a new **xPack Development Tools** project) would be prefered, toolchains with identical names but different functionality are confusing.

sequencer:
I also give this [issue](https://github.com/riscv/riscv-gcc/issues/157).
How a about change this packages’ prefix to `riscv-none-gnu`?

ilg-ul:
To my knowledge, GNU is an UNIX like operating system, not related with embedded/bare metal devices which `riscv-none-embed` addresses.

sequencer:
OK I'll close this PR, after finishing the patch work, I will reopen it.

sequencer:
@ilg-ul cherry picked unmerged commit from https://github.com/sifive/riscv-gcc.

sequencer:
Ping

ilg-ul:
Pong.

Sorry, but I cannot follow you, if you cherry pick individual commits, you are on your own, I don't know the details of SiFive development. In my release I used exactly the same sources as they said they did and compiled them as is. If I remember right, the only deviation was for GDB, which had some problems and I used a more recent commit.

ilg-ul:
Did you manage to integrate the toolchain in your distribution?

sequencer:
Currently I use `riscv64-linux-gnu-` as prefix with `-nostdinc` and `-nostdlib` for compiling. It's from the upstream.
Locally, I use upstream with local custom patch to create different packages for custom instruction integration.

PR #1231, opened 2019-07-03
[vnote-git]: fix pkgver

Rasphino:
`git describe` will ignore commit info if there is a tag points to this commit. Thus, we need to add a parameter `--long` to enforce `git describe` emits all the info we need.

And according to git's [document](https://git-scm.com/docs/git-describe#Documentation/git-describe.txt---long), `git describe --long` will describe such a commit as `v1.2-0-gdeadbee` (0th commit since tag v1.2 that points at object deadbee…​.). So we need to delete that `-0-` as well.


Xuanwo:
Please add package name into commit title.

PR #1305, opened 2019-09-22
Add package deepin-wine-tim

countstarlight:
Signed-off-by: Codist <countstarlight@gmail.com>


a-wing:
Please remove the useless files

countstarlight:
@a-wing I also thought that I needed to remove the useless files. But in keeping with github, these files have also been uploaded to aur, which may be pulled by lilacbot when pulling updates.

countstarlight:
Not get a reply for a long time, close.

PR #1327, opened 2019-10-13
rebuild lgogdownloader and let it update on boost

specter119:


lilydjwg:
There is no need to pkgrel+1 when `update_on` adds new entries. Also you need to update pkgrel when rebuild happens, e.g. using `update_pkgver_and_pkgrel(_G.newver.split('-')[0])` so users can see the update.

edward-p:
> no need to pkgrel

Got it.

Issue #1439, opened 2019-12-28
pid_children: dependency of lilac


yan12125:
### 问题类型 / Type of issues

* 软件打包请求 / package request

----

* 简述 / Brief Description

> A simple program outputs descendant process ids for each pid given on the command line

* 主要语言和工具链 / Main Languages and Toolchains

Rust

* 上游链接 / Upstream Link

https://github.com/lilydjwg/pid_children

* 为何需要这个软件包 / Why Need This Package

To make local installation of [lilac](https://github.com/archlinuxcn/lilac) easier.

Here is a PKGBUILD that makes lilac happy, but far from a good PKGBUILD:
```
pkgname=pid_children-git
pkgver=r9.d27761d
pkgrel=1
pkgdesc="A simple program outputs descendant process ids for each pid given on the command line"
arch=('x86_64')
depends=('gcc-libs')
makedepends=('git' 'rust')
conflicts=("pid_children")
provides=("pid_children=$pkgver")
url="https://github.com/lilydjwg/pid_children"
license=('custom:Unknown')
source=(git+https://github.com/lilydjwg/pid_children.git)
sha256sums=('SKIP')

pkgver() {
  cd pid_children
  printf "r%s.%s" "$(git rev-list --count HEAD)" "$(git rev-parse --short HEAD)"
}

build() {
  cd pid_children
  cargo build --release --locked
}

package() {
  cd pid_children
  install -Dm755 target/release/{kill_children,pid_children} -t "$pkgdir"/usr/bin
}
```

Some issues:
* Upstream does not declare a license
* What should be the owner/group and the permission of `kill_children`?

/cc @lilydjwg

lilydjwg:
`kill_children` should be 4754 owned by root:pkg.

lilydjwg:
License has been added to pid_children.

PR #1440, opened 2019-12-29
addpkg: pid_children-git

yan12125:
Closes: https://github.com/archlinuxcn/repo/issues/1439


archlinuxcn/pid_children-git/tmpfiles.conf
```diff
@@ -0,0 +1 @@
+z /usr/bin/kill_children - - pkg
```
lilydjwg: Shouldn't we specify permissions here? Changing group would discard the sgid bit.

yan12125: Thanks I forgot that is how chown/chgrp usually works. But again systemd does something different - it tries to keep suid/sgid as much as possible (https://github.com/systemd/systemd/blob/v244/src/tmpfiles/tmpfiles.c#L833). Anyway specifying permissions explicitly is better.

lilydjwg: What? It means that I can get root for free if I can find a systemd-managed root owned file that I can unlink and recreate....

yan12125:
Thanks for the comments!

PR #1488, opened 2020-02-11
coredns: Corefile: Use forward instead of proxy

outloudvi:
`proxy` is not compiled into coredns by default, `forward` does. For other info, see also Azure/AKS#1304.


Xuanwo:
Please run `updpkgsums` to update `PKGBUILD`'s sha256sums~

outloudvi:
Force-pushed and done.

Xuanwo:
Merged.

PR #1522, opened 2020-03-07
ferdi-git: switch to electron v8

jbbr:
Upstream ferdi is using electron v8 since [5.4.4-beta.2](https://github.com/getferdi/ferdi/releases/tag/v5.4.4-beta.2).

This package is still using electron v6. This PR changes that.


PeterCxy:
i switched to electron 6 because electron 8 started to cause segfaults since some time ago. I think ferdi has fixed it but I will keep using electron 6 for a while until ferdi has an actual release version with electron 8.

PR #1549, opened 2020-03-30
Better TimeoutStartSec=infinity instead of 0.

ideal:
See https://www.freedesktop.org/software/systemd/man/systemd.service.html


Xuanwo:
Submit to https://aur.archlinux.org/packages/k3s-bin/ instead.

Xuanwo:
Closed for not fixable from archlinuxcn side.

Issue #990, opened 2019-01-06
[RFC] Migrating build_prefix, depends and update_on to lilac.yaml


yan12125:
### 问题类型 / Type of issues

* 其它 / other

### 受影响的软件包 / Affected packages

Too many LOL

----

Settings in nvchecker.ini are migrated to lilac.yaml in #923. I'd like to push things fuurther: keep only pre_build and post_build in lilac.py and move all other fields to lilac.yaml. This simplifies static checking of packages (e.g. #979).

Relevant discussions future plans on IRC/Telegram:

```
[23:09:27] <tg2arch> [yan12125] 把剩下還在nvchecker.ini裡的包都搞定了，希望沒弄錯啥（
[23:13:39] <tg2arch> [依云] yan12125: 好耶～
[23:17:28] <tg2arch> [yan12125] 在考慮是否要把build_prefix depends update_on也都搬到lilac.yaml去，現在配置散落各處，覺得挺亂
[23:19:04] <tg2arch> [依云] yan12125: 搬，如果你有时间的话
[23:19:52] <tg2arch> [依云] 还有只有一句 fromlilaclib import * 和注释的 lilac.py 可以清空
[23:20:51] <tg2arch> [yan12125] 依云: 這樣的檔案還不少，看來得寫script了(
[23:21:27] <tg2arch> [依云] 都1155个包了啊
[23:21:38] <tg2arch> [yan12125] 沒內容的lilac.py想等lilac不再需要空的lilac.py時直接刪掉(
[23:22:13] <tg2arch> [依云] 我还没想清楚怎么去掉 lilac.py
[23:23:06] <tg2arch> [yan12125] 是判斷一個包是否managed那邊不容易處理嗎？
[23:23:16] <tg2arch> [依云] yan12125: 嗯
[23:24:23] <tg2arch> [yan12125] 恩這真的挺麻煩的，有些包有lilac.yaml，可是卻不是managed...
[23:32:32] <tg2arch> [依云] yan12125: 要给那些手工管理的包加个 managed: false
[23:34:10] <tg2arch> [yan12125] 依云: 這樣應該可行～寫搬遷script的時候順便來加上
[23:35:02] <tg2arch> [依云] yan12125: 然后 lilac 那边的逻辑要改成先加载所有 yaml，再加载部分 lilac.py 并合并
[23:40:29] <tg2arch> [yan12125] 依云: 瞭 這個做完才能開始刪lilac.py對吧？
[...]

PR #1574, opened 2020-04-11
Test for invalid repo_depends items (closes #1414)

### LINKED ISSUE #1414: Invalid repo_depends
### 问题类型 / Type of issues

* 打包错误 / packaging error

### 受影响的软件包 / Affected packages

- [ ] archcn-tools-meta
- [ ] bilidan-git
- [ ] fs-uae-arcade-devel
- [ ] lib32-qt4
- [ ] libbonoboui
- [ ] libgnome
- [ ] libgnomeui
- [ ] pyenv-virtualenv
- [ ] ruby-globalid

----

Following-up of https://github.com/archlinuxcn/repo/issues/1351#issuecomment-560205352.

Some packages have non-existent or unmanaged pkgbase in repo_depends. For example, libgnomeui depends on non-existent gnome-vfs-nosmb, and archcn-tools-meta depends on unmanaged lilacinfo.

Here is the full list.

```
archcn-tools-meta depends on lilacinfo, which does not exist or is unmanaged
bilidan-git depends on danmaku2ass-git, which does not exist or is unmanaged
fs-uae-arcade-devel depends on python-lhafile, which does not exist or is unmanaged
lib32-qt4 depends on lib32-libmng, which does not exist or is unmanaged
libbonoboui depends on gnome-vfs-nosmb, which does not exist or is unmanaged
libgnome depends on gnome-vfs-nosmb, which does not exist or is unmanaged
libgnomeui depends on gnome-vfs-nosmb, which does not exist or is unmanaged
pyenv-virtualenv depends on pyenv, which does not exist or is unmanaged
ruby-globalid depends on ruby-activesupport, which does not exist or is unmanaged
```
The list is generated from the following script:
```Python
import pathlib
import sys

sys.path.append('vendor')

from lilac2.lilacpy import load_managed

mods, errors = load_managed(pathlib.Path('../repo/archlinuxcn').resolve())
assert not errors
for name, mod in mods.items():
    for dep in getattr(mod, 'repo_depends', []):
        if isinstance(dep, tuple):
            pkgbase = dep[0]
        else:
            pkgbase = dep
        if pkgbase not in mods:
            print(f'{name} depends on {pkgbase}, which does not exist or is unmanaged')
```
Such problems might need to be checked periodically https://github.com/archlinuxcn/misc_scripts/issues/2.

yan12125:
Marked as draft as this depends on https://github.com/archlinuxcn/lilac/pull/144. Will update after that pull request is merged or another approach is proposed.

See https://travis-ci.org/github/yan12125/repo/builds/673671061 for an example failing test.

### DIFF
--- a/pre-commit
+++ b/pre-commit
@@ -372,6 +372,27 @@ class RepoTreeTest(TestCase):
                     check_var_srcinfo('replaces', package, PACKAGES, line)
                     check_var_srcinfo('groups', package, GROUPS, line)
 
+    def test_repo_depends_valid(self):
+        pkgs = list(list_packages())
+        for package in pkgs:
+            with self.subTest(package=package):
+                self.assertFalse(
+                    lilac_py_has_field(package, 'repo_depends', self),
+                    msg=('package "\033[1;31m{0}\033[m" has repo_depends '
+                         'in lilac.py').format(package))
+                lilac_yaml_path = os.path.join(pkgpath(package), 'lilac.yaml')
+                with git_open(lilac_yaml_path) as lilac_yaml_file:
+                    lilac_yaml = yaml.load(lilac_yaml_file.read(), Loader=yaml.SafeLoader)
+                    for dep in lilac_yaml.get('repo_depends', []):
+                        if isinstance(dep, dict):
+                            dep_package = list(dep.keys())[0]
+                        else:
+                            dep_package = dep
+                        self.assertTrue(
+                            dep_package in pkgs,
+                            msg=('package "\033[1;31m{0}\033[m" depends on '
+                                 '"\033[1;31m{1}\033[m", which does not exist '
+                                 'or is unmanaged').format(package, dep_package))
 
 
 def run_test(testcase, msg):
--- a/pre-commit
+++ b/pre-commit
@@ -385,6 +385,10 @@ class RepoTreeTest(TestCase):
                     lilac_yaml = yaml.load(lilac_yaml_file.read(), Loader=yaml.SafeLoader)
                     for dep in lilac_yaml.get('repo_depends', []):
                         if isinstance(dep, dict):
+                            # When `dep` is a dict, this repo_depends entry
+                            # has the form `pkgbase: pkgname`. Here I assume
+                            # split packages are successfully packaged. so only
+                            # the pkgbase needs to be checked.
                             dep_package = list(dep.keys())[0]
                         else:
                             dep_package = dep
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 
 python:
-    - "3.6"
+    - "3.8"
 
 # Use container-based infrastructure for quicker build start-up
 sudo: false
--- a/archlinuxcn/armadillo/lilac.yaml
+++ b/archlinuxcn/armadillo/lilac.yaml
@@ -11,6 +11,3 @@ post_build: aur_post_build
 
 update_on:
   - aur: armadillo
-
-repo_depends:
-  - superlu
--- a/archlinuxcn/fs-uae-arcade-devel/lilac.yaml
+++ b/archlinuxcn/fs-uae-arcade-devel/lilac.yaml
@@ -8,7 +8,6 @@ build_prefix: extra-x86_64
 
 repo_depends:
   - fs-uae-devel
-  - python-lhafile
 
 pre_build: aur_pre_build
 


yan12125:
TODO if this PR is accepted: reject in-official-repos and orphaning issues in issuebot [1] if the to-be-removed package is listed in repo_depends of another package.

[1] https://github.com/archlinuxcn/misc_scripts/blob/master/issuebot

lilydjwg:
To read `repo_depends` without `ImportError`s you can load only `lilac.yaml`, or hack the importing system to return fake modules and hope it works.

yan12125:
> you can load only `lilac.yaml`

For now it does not work as

```
$ rg -lg lilac.py repo_depends archlinuxcn | wc -l
23
```
And as you know, not everyone agreed on migrating `repo_depends` to `lilac.yaml` (#990).

> hack the importing system to return fake modules

Do you mean something like the commit 77660f3e990fad5f477ddaf9614556fd8fa25aac in #979? I'm fine if such a commit is the only stuff to add. A bigger issue is that one needs to replicate complex logic to handle `lilac.py` and `lilac.yaml`, like `_parse_lilac`, `_parse_lilac_py` and `_parse_lilac_yaml` functions in #979, or `lilacpy.py` and `lilacyaml.py` in lilac2. I'm afraid that will be a big maintenance burden.

I guess you want to make the new test run no matter a packager clones lilac or not? An approach might be vendoring necessary files from lilac2 and winterpy and adjust `sys.path` in pre-commit. I can create a script for that if that sound fine to you.

lilydjwg:
On Sun, Apr 12, 2020 at 01:17:48AM -0700, Chih-Hsuan Yen wrote:
> > you can load only `lilac.yaml`
> 
> For now it does not work as
> 
> ```
> $ rg -lg lilac.py repo_depends archlinuxcn | wc -l
> 23
> ```
> And as you know, not everyone agreed on migrating `repo_depends` to `lilac.yaml` (#990).

Sigh... We should port them. lilac.yaml is there to enable wider and
safer scripting, which is exactly what a check script needs.

> > hack the importing system to return fake modules
> 
> Do you mean something like the commit 77660f3e990fad5f477ddaf9614556fd8fa25aac in #979? I'm fine if such a commit is the only stuff to add. A bigger issue is that one needs to replicate complex logic to handle `lilac.py` and `lilac.yaml`, like `_parse_lilac`, `_parse_lilac_py` and `_parse_lilac_yaml` functions in #979, or `lilacpy.py` and `lilacyaml.py` in lilac2. I'm afraid that will be a big maintenance burden.

3rd party modules are usually used inside functions so if we don't run
the functions, and provide a fake module that can be imported, it should
work. We can hook into the importing system and turn every unknown
module to a fake one.

But this is not a good way to go.

> I guess you want to make the new test run no matter a packager clones lilac or not? An approach might be vendoring necessary files from lilac2 and winterpy and adjust `sys.path` in pre-commit. I can create a script for that if that sound fine to you.

No, we should "vendor" the needed data.

In fact, sometimes I am thinking about move more data from PKGBUILD into
a declarative representation. That's to say, use a TOML file to specify
package name, dependencies, source URLs, etc, so we could have more
metadata to use, fast, safely and without bashes.

-- 
Best regards,
lilydjwg

yan12125:
So, let's go back to #990 first?

yan12125:
Rebased after #1615. I also added a check to deny `repo_depends` in lilac.py

yan12125:
Rebased and tested manually. A new broken package `armadillo` is now fixed.

Given that in-official-repos and orphaned packages are now automatically handled by the issuebot, could this be merged?

pre-commit
```diff
@@ -372,6 +372,27 @@ class RepoTreeTest(TestCase):
                     check_var_srcinfo('replaces', package, PACKAGES, line)
                     check_var_srcinfo('groups', package, GROUPS, line)
 
+    def test_repo_depends_valid(self):
+        pkgs = list(list_packages())
```
lilydjwg, line 376: This looks strange. `list_packages` should be called `iter_packages`.

yan12125: Sort of. `list_packages` has been a generator since its introduction.

pre-commit
```diff
@@ -372,6 +372,27 @@ class RepoTreeTest(TestCase):
...
+                self.assertFalse(
+                    lilac_py_has_field(package, 'repo_depends', self),
+                    msg=('package "\033[1;31m{0}\033[m" has repo_depends '
+                         'in lilac.py').format(package))
+                lilac_yaml_path = os.path.join(pkgpath(package), 'lilac.yaml')
+                with git_open(lilac_yaml_path) as lilac_yaml_file:
+                    lilac_yaml = yaml.load(lilac_yaml_file.read(), Loader=yaml.SafeLoader)
+                    for dep in lilac_yaml.get('repo_depends', []):
+                        if isinstance(dep, dict):
+                            dep_package = list(dep.keys())[0]
```
farseerfc, line 388: My understanding of reading this line is that, this only checks pkgbase name for splitpkgs, which is fine for solving #1414 and should work in most cases (the number of `repo_depends: - pkgbase: pkgname` is limited). Maybe we should leave a note about this for future improvements.

pre-commit
farseerfc: we can change this function name after this got merged

farseerfc:
LGTM except for the note on splitpkgs name.

pre-commit
yan12125: Thanks! I added a not about this line in the new commit.

farseerfc:
Should I merge it and make the change?

PR #1614, opened 2020-05-09
chromium-vaapi: Should rebuild on "re2" update

outloudvi:
Recent `re2` update (`1:20200401-1` -> `1:20200501-1`) breaks `chromium-vaapi`:

```
/usr/lib/chromium/chromium: error while loading shared libraries: libre2.so.6: cannot open shared object file: No such file or directory
```

I don't really know if it's how lilac works, but maybe we can rebuild `chromium-vaapi` on `re2` upgrades.


MarvelousBlack:
https://github.com/archlinuxcn/repo/commit/549a4b753262c7057b0602e49efc3cd7cfe0b3e1

PR #1655, opened 2020-06-02
qv2ray: rebuild due to protobuf soversion bump

DuckSoft:
upstream `protobuf` bumped soversion recently.
a rebuilld is needed.


lilydjwg:
why not use update_on field? there is an alias for protobuf already.

DuckSoft:
> why not use update_on field? there is an alias for protobuf already.

Thanks! I'll add that for qv2ray now, and qv2ray-dev-git afterwards.

DuckSoft:
@lilydjwg ready to be merged

PR #1658, opened 2020-06-06
addpkg: trojan-go-git

DuckSoft:
https://github.com/p4gefau1t/trojan-go

PR #1659, opened 2020-06-06
qv2ray-dev-git: delete lilac.py

DuckSoft:
removed since we should not update AUR if it's a VCS package


yuyichao:
Doesn't this affect update of manual update as well?

yuyichao:
And I believe liliac is no longer pushing any commit that is violating the AUR convention anymore.

lilydjwg:
It's OK to use `update_aur_repo` for vcs packages now, because it will no longer push useless commits.

yuyichao:
Also ref https://github.com/archlinuxcn/repo/pull/1658#discussion_r436277351 which I think should also be kept unless the PKGBUILD will never be editted manually or if the AUR should not be kept in sync.

DuckSoft:
Okay. Closing this.

PR #1701, opened 2020-06-28
pre-commit: fix repo_depends check

yan12125:
When pre-commit is used as a git hook, iter_packages only goes through
changed packages, and dependent package may not be in `pkgs`.

Test:
1. Modify random stuffs in libqtxdg-git
2. Commit libqtxdg-git

Follow-up of #1574.

### DIFF
--- a/pre-commit
+++ b/pre-commit
@@ -24,6 +24,7 @@ from time import strftime
 from subprocess import Popen, PIPE
 from unittest import TestCase, TestLoader, TextTestRunner, skip
 from io import StringIO
+from typing import Generator
 
 # need python-yaml package
 import yaml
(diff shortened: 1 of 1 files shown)


pre-commit
```diff
@@ -272,7 +272,7 @@ def load_lilac_yaml_schema():
         YAML_SCHEMA = yaml.load(schema, Loader=yaml.SafeLoader)
 
 
-def iter_packages():
+def iter_packages(all_pkgs: bool = False):
```
lilydjwg, line 275: Return type not specified; mypy will skip.

yan12125: Thanks I often forget return types :sweat_smile:  Fixed now.

PR #1888, opened 2020-09-08
mesen: ArchLinuxCN init

CrankySupertoon:


archlinuxcn/mesen/PKGBUILD
```diff
@@ -0,0 +1,51 @@
+# Maintainer:  Andrew O'Neill <andrew at meanjollies dot com>
+# Contributor: John Jenkins <twodopeshaggy@gmail.com>
+
+pkgname=mesen
+_pkgname=Mesen
+pkgver=0.9.9
+pkgrel=1
+pkgdesc='A cross-platform NES/Famicom emulator'
+arch=('x86_64')
+makedepends=('clang' 'gendesk' 'zip' 'msbuild-stable')
```
dctxmei, line 10:
```suggestion
makedepends=('clang' 'gendesk' 'zip' 'mono-msbuild')
```

1. The msbuild-stable package has disappeared, which will cause the build to fail.
2. I modified it to the mono-msbuild package and successfully completed the build, but I did not test it.
3. If you think this change is reasonable, you also need to add a lilac.py file to ensure that this content will be modified after the file is pulled from AUR.
4. Another option is not to use AUR as the upstream, but to maintain it separately in Arch Linux CN Repo.

dctxmei (changes requested):

archlinuxcn/mesen/lilac.yaml
```diff
@@ -0,0 +1,12 @@
+maintainers:
+  - github: meanjollies
```
dctxmei, line 2:
```suggestion
  - github: CrankySupertoon
```

You are the maintainer, not someone else. It has nothing to do with who the maintainer on AUR is.

dctxmei (changes requested):

dctxmei:
It takes too long.

PR #1906, opened 2020-09-18
CI: Add GitHub Actions

imlonghao:
Add Github Actions support, doing the same job as travis does, but with a better GitHub integrated support.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,17 @@
+name: Packaging consistency check
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,17 @@
...
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.8'
+    - name: Install dependencies
+      run: pip install pyaml jsonschema
```
lilydjwg, line 15: Do we need pyaml instead of PyYAML?

imlonghao: Looks like pyaml is built on top of PyYAML, we can use PyYAML instead.

PR #1972, opened 2020-11-13
addpkg: xf86-input-libinput-git-oldherl

oldherl:


oldherl:
Using technique similar to qt5-base-24h.

oldherl:
Found a problem: PKGBUILD is not downloaded each time from AUR before building

PR #2035, opened 2020-12-25
addpkg: haskell-json-autotype

berberman:


archlinuxcn/haskell-json-autotype/lilac.yaml
```diff
@@ -0,0 +1,36 @@
+maintainers:
+  - github: berberman
+
+repo_depends:
+  - haskell-genericpretty
+  - haskell-json-alt
+  - haskell-run-haskell-module
```
dctxmei, line 7: These three packages do not exist.

PR #2039, opened 2020-12-31
addpkg: lilac-git

yan12125:
Depends on ongoing upstream PRs: https://github.com/archlinuxcn/lilac/pull/170, https://github.com/archlinuxcn/lilac/pull/171

@lilydjwg I bundled [build-cleaner](https://github.com/archlinuxcn/misc_scripts/blob/master/build-cleaner) in this package as it seems the simplest way. As you're the only contributor for that script, could you go ahead and declare a license for it?


lilydjwg:
I'm going to move build-cleaner into lilac's repo since it's now required.

yan12125:
This should be ready to go. Mind to have a look?

PR #2103, opened 2021-02-21
llvm-git and more "modern" lilac style i think

SolarAquarion:


lilydjwg:
@yuyichao do you want to take a look?

archlinuxcn/llvm-git/PKGBUILD
```diff
@@ -0,0 +1,404 @@
...
+
+    rm -rf "$srcdir"/fakeinstall*
+    
+    export _ocamlver=$(ocamlc --version)
+    export _pythonver=$(python -V | sed 's/.* \([0-9]\).\([0-9]\).*/\1.\2/')
+}
+
+build() {
+    export CFLAGS="$(echo "$CFLAGS" | sed -e "s/-fstack-protector-strong//") -I/usr/include/tensorflow "
+    export CXXFLAGS="$(echo "$CXXFLAGS" | sed -e "s/-fstack-protector-strong//") -I/usr/include/tensorflow "
```
yuyichao, line 72: tensorflow isn't added as makedepends

archlinuxcn/llvm-git/PKGBUILD
```diff
@@ -0,0 +1,404 @@
...
+
+
+# Based on LordHeavy's LLVM
+pkgbase=llvm-git
+pkgname=('lldb-git' 'lld-git' 'polly-git' 'compiler-rt-git' 'clang-git' 'llvm-ocaml-git' 'llvm-libs-git' 'llvm-git'  'lib32-llvm-libs-git' 'lib32-clang-git' 'lib32-llvm-git') 
+pkgdesc='Low Level Virtual Machine (git version)'
+pkgver=12.0.0_r376876.6f0f0220380f
+pkgrel=1
+groups=('chaotic-mesa-git')
+arch=('x86_64')
```
yuyichao, line 11: Please include armv7h and aarch64

archlinuxcn/llvm-git/lilac.yaml
```diff
@@ -0,0 +1,10 @@
+maintainers:
+  - github: solaraquarion
+
+build_prefix: extra-x86_64
+
+pre_build: vcs_update
+
+update_on:
```
yuyichao, line 8: Should update on ocaml and python as well.

See `package_llvm-ocaml-julia` from `llvm-julia` for how you can add ocaml version dependency automatically.

archlinuxcn/llvm-git/lilac.yaml
```diff
@@ -0,0 +1,10 @@
+maintainers:
+  - github: solaraquarion
+
+build_prefix: extra-x86_64
+
+pre_build: vcs_update
```
yuyichao, line 6: `post_build: git_pkgbuild_commit`, no need for `lilac.yaml`

archlinuxcn/llvm-git/PKGBUILD
yuyichao: Would also be nice if you can turn `-fasynchronous-unwind-tables` on.

archlinuxcn/llvm-git/PKGBUILD
```diff
@@ -0,0 +1,404 @@
...
+               'lib32-libxml2' 'lib32-zlib' 'python2' ) #tensorflow
+
+source=("llvm-project::git+https://github.com/llvm/llvm-project.git"
+        "llvm-config.h")
+
+md5sums=('SKIP'
+         '295c343dcd457dc534662f011d7cff1a')
+sha512sums=('SKIP'
+            '75e743dea28b280943b3cc7f8bbb871b57d110a7f2b9da2e6845c1c36bf170dd883fca54e463f5f49e0c3effe07fbd0db0f8cf5a12a2469d3f792af21a73fcdd')
+options=('staticlibs')
```
yuyichao, line 27: Could you also include `debug` please?

archlinuxcn/llvm-git/PKGBUILD
```diff
@@ -0,0 +1,404 @@
...
+       done
+}
+
+package_lldb-git() {
+     pkgdesc="Next generation, high-performance debugger (git version)"
+     url="https://lldb.llvm.org/"
+     depends=("clang-git=$pkgver-$pkgrel" "llvm-libs-git=$pkgver-$pkgrel")
+     optdepends=('llvm-git: for the man file')
+     provides=("lldb=$pkgver")
+     conflicts=('lldb')
```
yuyichao, line 172: Missing replaces and provides for the svn version? (or have they never existed before?)

Same for `lld`, `lib32-clang`, `lib32-llvm` below (I feel like if `lib32-llvm-libs-svn` existed `lib32-llvm-svn` probably did as well...)

archlinuxcn/llvm-git/PKGBUILD
```diff
@@ -0,0 +1,404 @@
...
+  install -Dm644 "$srcdir"/llvm-project/clang/LICENSE.TXT "$pkgdir/usr/share/licenses/$pkgname/LICENSE"
+}
+
+package_clang-git() {
+     pkgdesc="C language family frontend for LLVM (git version)"
+     url="http://clang.llvm.org/"
+     depends=("llvm-git=$pkgver-$pkgrel" 'gcc' "python" 'python2')
+     optdepends=('llvm-libs: for compiling with -flto')
+     provides=("clang" 'clang-analyzer' 'clang-tools-extra')
+     conflicts=('clang' 'clang-svn' 'clang-analyzer' 'clang-tools-extra' 'clang-tools-extra-svn')
```
yuyichao, line 233: Missing `clang-analyzer-svn` conflict.

archlinuxcn/llvm-git/PKGBUILD
```diff
@@ -0,0 +1,404 @@
...
+     replaces=('clang-svn' 'clang-analyzer-svn' 'clang-tools-extra-svn' 'clang')
```
yuyichao, line 234: **Do not** replace `clang` or any other non-vcs packages.

Same below for: `llvm-libs`, `llvm-ocaml`, `lib32-llvm-libs`, `lib32-clang`, `lib32-llvm`.

archlinuxcn/llvm-git/PKGBUILD
```diff
@@ -0,0 +1,404 @@
...
+     _fakeinstall_64 fakeinstall_64/usr/lib/python"$_pythonver"/site-packages/lldb 
+     _fakeinstall_64 fakeinstall_64/usr/share/man/man1/lldb*
+    install -Dm644 "$srcdir"/llvm-project/lldb/LICENSE.TXT "$pkgdir/usr/share/licenses/$pkgname/LICENSE"
+
+}
+
+package_lld-git(){
+   pkgdesc="Linker from the LLVM project"
+   url="https://lld.llvm.org"
+   depends=("gcc-libs")
```
yuyichao, line 185: Missing `llvm-libs-git` dependency (you aren't statically linking LLVM are you?)

Version dependency for `llvm-libs-git` also missing for `lib32-clang-git` below.

archlinuxcn/llvm-git/PKGBUILD
```diff
@@ -0,0 +1,404 @@
...
+        -D FFI_INCLUDE_DIR="$(pkg-config --variable=includedir libffi)" \
+        -D LLVM_BINUTILS_INCDIR=/usr/include \
+        -D LLVM_APPEND_VC_REV=ON 
+
+    ninja -C _build32 all ocaml_doc LLVMgold
+    DESTDIR="$srcdir/fakeinstall_32" ninja -C _build32 install
+}
+
+_fakeinstall_64 () {
+        local src f dir
```
yuyichao, line 147: Very minor, but indentation is a bit strange here.... Also the case for the 32 version and some loops, `local` below....

yuyichao (changes requested):
I have a local WIP on this as well which I was going to finish when I have time/need for it .... I'm more than happy if someone else want to take this though.

I did not do a test compilation or check if the result of the compilation works. The comments I made are basically all about metadata and some compilation options that are/were useful to me and shouldn't really affect / should help other users as well.

archlinuxcn/llvm-git/PKGBUILD
yuyichao: haha, I guess the CI caught some of these at least =)

archlinuxcn/llvm-git/PKGBUILD
```diff
@@ -0,0 +1,404 @@
...
+    rm -rf debuginfo-tests libclc libcxx libcxxabi llgo openmp parallel-libs pstl libc
+
+    rm -rf "$srcdir"/fakeinstall*
+    
+    export _ocamlver=$(ocamlc --version)
+    export _pythonver=$(python -V | sed 's/.* \([0-9]\).\([0-9]\).*/\1.\2/')
+}
+
+build() {
+    export CFLAGS="$(echo "$CFLAGS" | sed -e "s/-fstack-protector-strong//") -I/usr/include/tensorflow "
```
yuyichao, line 71: And just curious, did the stack protector break something or was this just to get better performance?

I'm also pretty sure that it'll be more reliable if you add a `-fno-stack-protector` before/after instead. Had arch changes the default protection option this version will most likely break due to string mismatch.

archlinuxcn/llvm-git/PKGBUILD
yuyichao: And in that case the `lib32` package should be conditional on `$CARCH = x86_64`.

archlinuxcn/llvm-git/PKGBUILD
```diff
@@ -0,0 +1,404 @@
...
+    -B _build32 \
+    -S llvm-project/llvm \
+    -G Ninja \
+        -D LLVM_ENABLE_PROJECTS="clang;clang-tools-extra;compiler-rt" \
+        -D CMAKE_BUILD_TYPE=Release \
+        -D CMAKE_INSTALL_PREFIX=/usr \
+        -D LLVM_LIBDIR_SUFFIX=32 \
+        -D CMAKE_C_FLAGS="$LIB32_CFLAGS" \
+        -D CMAKE_CXX_FLAGS="$LIB32_CXXFLAGS" \
+        -D LLVM_TARGET_ARCH:STRING=i686 \
```
yuyichao, line 126: I guess this target is all what you need for the lib32 version?

archlinuxcn/llvm-git/lilac.yaml
SolarAquarion: i did a bunch of change. hopefully it's better

SolarAquarion:
@yuyichao basically it's a monolithic package set, which means it's supposed to work for x86_64 and whatever.  So i don't know how to particularly like segregate those things

yuyichao:
> basically it's a monolithic package set, which means it's supposed to work for x86_64 and whatever

No not really, in fact the PKGBUILD already had condition on architectures.
See the changes I just pushed for how it should be done.

A few things to note,

1. Please don't set variables in `prepare`, or any special functions FWIW that are used by other functions. They will not work when running only certain steps (e.g. as what it was, `makepkg -R` won't work).
2. `CFLAGS`, `CXXFLAGS`, `LDFLAGS` stores a space separated list of strings, make sure you add a space when doing `+=`.
3. I don't think `polly` should have "any" arch, since it does install files under `/usr/lib` and I'm 99% sure some of the files there are libraries.

I also fixed lilac and the format. The commit should probably be squashed before merging.

@lilydjwg IIRC you use the package_* functions to figure out which package a PKGBUILD make create? So I shouldn't need to provide the package name list separately right? (I'm 99% sure since I've been doing something similar for years)

archlinuxcn/llvm-git/PKGBUILD
```diff
@@ -0,0 +1,421 @@
...
+        -D CLANG_INCLUDE_DOCS=ON \
+        -D LLVM_EXPERIMENTAL_TARGETS_TO_BUILD=AVR \
+        -D LLVM_ENABLE_SPHINX=ON \
+        -D SPHINX_OUTPUT_HTML=OFF \
+        -D SPHINX_WARNINGS_AS_ERRORS=OFF \
+        -D POLLY_ENABLE_GPGPU_CODEGEN=ON \
+        -D LLVM_VERSION_SUFFIX="" \
+        -D LLDB_USE_SYSTEM_SIX=1 \
+        -D LLVM_ENABLE_PROJECTS="lldb;polly;compiler-rt;lld;clang-tools-extra;clang" \
+        -D TENSORFLOW_C_LIB_PATH="/usr/include/tensorflow/"
```
yuyichao, line 109: I'm doing a test compilation right now so I might be able to answer this question myself in a few hours, but does this create a runtime dependency on tensorflow, or is that purely compile time?

yuyichao: It's also a bit strange that you are passing a include path to something called "lib_path"...

yuyichao:
Another note that you'll probably need to package [python-sphinx-automodapi](https://aur.archlinux.org/packages/python-sphinx-automodapi/) to get the compilation to pass due to https://github.com/llvm/llvm-project/commit/e7bc6c594b75602c23cb901f53b3a30d48e2ee78

yuyichao (approved):

yuyichao (changes requested):
My previous comments should be all fixed. However, new dependency is now necessary and I probably won't be doing **that** myself very soon...

yuyichao:
> However, new dependency is now necessary and I probably won't be doing that myself very soon...

Especially since the lldb 12 package will probably need it and it'll hopefully be pulled into the official repo at that time....

SolarAquarion:
Just disable Sphinx for now i guess?

yuyichao:
> Just disable Sphinx for now i guess?

That's fine by me.

SolarAquarion:
Sphinx has already been disabled.

archlinuxcn/llvm-git/PKGBUILD
```diff
@@ -0,0 +1,421 @@
...
+        -D FFI_INCLUDE_DIR="$(pkg-config --variable=includedir libffi )" \
+        -D LLVM_BUILD_LLVM_DYLIB=ON \
+        -D LLVM_LINK_LLVM_DYLIB=ON \
+        -D LLVM_INSTALL_UTILS=ON \
+        -D LLVM_BUILD_DOCS=ON \
+        -D LLVM_INCLUDE_DOCS=ON \
+        -D LLVM_ENABLE_DOXYGEN=OFF \
+        -D CLANG_INCLUDE_DOCS=ON \
+        -D LLVM_EXPERIMENTAL_TARGETS_TO_BUILD=AVR \
+        -D LLVM_ENABLE_SPHINX=ON \
```
yuyichao, line 102: I don't think so....

SolarAquarion:
It's disabled on lib32, not on the main ones

archlinuxcn/llvm-git/PKGBUILD
```diff
@@ -0,0 +1,421 @@
...
+        -D LLVM_VERSION_SUFFIX="" \
+        -D LLDB_USE_SYSTEM_SIX=1 \
+        -D LLVM_ENABLE_PROJECTS="lldb;polly;compiler-rt;lld;clang-tools-extra;clang" \
+        -D TENSORFLOW_C_LIB_PATH="/usr/" # Force on
+
+    ninja -C _build64 LLVMgold  all ocaml_doc
+    DESTDIR="$srcdir/fakeinstall_64" ninja -C _build64 install
+
+    if [[ $CARCH == x86_64 ]]; then
+        export PKG_CONFIG_PATH="/usr/lb32/pkgconfig"
```
lilydjwg, line 115: `lb32` ?

yuyichao:
> not on the main ones

Right, so the compilation will fail.

SolarAquarion:
Disabled Sphinx, I think.

yuyichao:
I highly suspect that's not enough and it may error in the package function...

SolarAquarion:
this should be better now?

yuyichao:
A few more changes to make it more friendly to the build machine/other packages.

1. `-debug` packages take way too long to create (a few hours at least) so I disabled it...
2. The lib32 package doesn't have to be built at the same time AFAICT. Doing so exhausts the disk space on the tmpfs on the build machine so it's separated out.
3. Change update frequency to every two days instead of following the LLVM git update, which will probably be 3 times a day. I don't think that update frequency is needed....

I'm testing compilation again on the build server to see if things can finish within the quota now... If not this will need to wait for some server configuration update first...

SolarAquarion:
Daily, or Every Other Day or so is good enough.

SolarAquarion:
what the fuck did i do now.  Did a push that  I do get fucked up?

yuyichao:
you reverted everything to an old commit.

SolarAquarion:
uh, yeah, my branch is pretty old.  I think

yuyichao:
What change were you trying to make? If you haven't made any, there should not be any reason to push.

SolarAquarion:
I'm honestly just trying to sync my local one, with the one you mde.

yuyichao:
you should reset your local branch to the remote one. In any case, if you are trying to sync your local change, there should be no reason to push.

SolarAquarion:
aka git reset --hard, which i just did

yuyichao (approved):

PR #2106, opened 2021-02-23
llvm-git: adding dependencies

SolarAquarion:
@yuyichao


archlinuxcn/llvm-git/PKGBUILD
```diff
@@ -121,7 +121,7 @@ package_lldb-git() {
 
     pkgdesc="Next generation, high-performance debugger (git version)"
     url="https://lldb.llvm.org/"
-    depends=("clang-git=$pkgver-$pkgrel" "llvm-libs-git=$pkgver-$pkgrel")
+    depends=("clang-git=$pkgver-$pkgrel" "llvm-libs-git=$pkgver-$pkgrel" "python" "libedit" "libxml2")
```
yuyichao, line 124: The official `lldb` do not have `libedit` and `libxml2` as dependency and they are already dependency of `llvm-libs-git`. Are these just to make sure all directly linked packages are listed as dependencies? (It's not arch's official policy though I'm good with that.)

archlinuxcn/llvm-git/PKGBUILD
```diff
@@ -155,7 +155,7 @@ package_lld-git(){
 package_polly-git() {
     pkgdesc="Polly is a high-level loop and data-locality optimizer and optimization infrastructure for LLVM"
     url="https://polly.llvm.org/"
-    depends=("llvm-git=$pkgver-$pkgrel")
+    depends=("llvm-libs-git=$pkgver-$pkgrel" "gcc-libs")
```
yuyichao, line 158: *I think* gcc-libs is typically not listed as dependency..... Or almost everything would have it...

I guess I'm ok with adding it but then it should at least be consistent.....

archlinuxcn/llvm-git/PKGBUILD
```diff
@@ -172,6 +172,7 @@ package_compiler-rt-git() {
     pkgdesc="Compiler runtime libraries for clang (git version)"
     url="https://compiler-rt.llvm.org/"
     depends=("llvm-git=$pkgver-$pkgrel" 'gcc-libs')
+    depends_x86_64=('lib32-gcc-libs')
```
yuyichao, line 175: Is this really a dependency? Or is it just a optional one. I assume you don't need the lib32 version if you are just using the x64 libraries.

archlinuxcn/llvm-git/PKGBUILD
```diff
@@ -190,7 +191,7 @@ package_clang-git() {
 
     pkgdesc="C language family frontend for LLVM (git version)"
     url="http://clang.llvm.org/"
-    depends=("llvm-git=$pkgver-$pkgrel" 'gcc' "python" 'python2')
+    depends=("llvm-git=$pkgver-$pkgrel" 'gcc' "python" 'python2' 'perl')
```
yuyichao, line 194: Optional? Do every command in this package uses it? (I kinda doubt `clang` do at the least). The official clang package does not seem to have `perl` as dependency or optional dependency. The python/python2 dependency should also probably be removed if not made optional.

yuyichao: And looking at the optional dependency below, I think `llvm-libs-git=$pkgver-$pkgrel` should be a dependency whereas `llvm-git` should be an optional one....

yuyichao:
@SolarAquarion As an aside, seems that lilac isn't happy that you are the maintainer without a public email address. @lilydjwg and others probably knows better what are all the options.....

SolarAquarion:
I'm just doing what namcap says I should add.  Technically not everything
depends on gcc-libs.  Like namcap, doesn't have linkage stuff

yuyichao:
> I'm just doing what namcap says I should add.

namcap is pretty bad.... It recommends quite a few anti-pattern AFAICT....

> Technically not everything depends on gcc-libs.  Like namcap, doesn't have linkage stuff

... Well, not everything depend on glibc for the same reasoning...... Every C++ program/library do though yet few if any of them adds it as dependency for this reason... Anyway, it should either be added consistently to everything within this package, or there should be something special about this particular package that requires adding the dependency....

archlinuxcn/llvm-git/PKGBUILD
SolarAquarion: It seemingly asks for it in namcap.

SolarAquarion:
Whatever I guess? I like not seeing namcap stuff, unfortunately, optdepends
will always be sort of a nuisance.

yuyichao:
Adding true unnecessary dependency (i.e. lib32) is bad for the user of the package so you must not do that.
For other ones that doesn't matter that much it's ok but it should be done consistently.

And you should just ignore namcap. It's not doing any good...

archlinuxcn/llvm-git/PKGBUILD
```diff
@@ -190,8 +191,12 @@ package_clang-git() {
 
     pkgdesc="C language family frontend for LLVM (git version)"
     url="http://clang.llvm.org/"
-    depends=("llvm-git=$pkgver-$pkgrel" 'gcc' "python" 'python2')
-    optdepends=('llvm-libs: for compiling with -flto')
+    depends=("llvm-libs-git=$pkgver-$pkgrel" 'gcc')
+    optdepends=('llvm-libs: for compiling with -flto'
```
yuyichao, line 195: llvm-libs shouldn't be optdepends anymore.
You can copy the note from the official `clang` for `llvm-git` and `python`. I assume python2 is for the scripts listed below it's probably fine to have it without a note. For `perl` if namcap recommended that it should probably have told you why it's needed and you should add that in the note.

SolarAquarion:
lldb-git E: Dependency libedit detected and not included (libraries ['usr/lib/libedit.so.0'] needed in files ['usr/lib/liblldb.so.13.0.0'])
lldb-git E: Dependency libxml2 detected and not included (libraries ['usr/lib/libxml2.so.2'] needed in files ['usr/lib/liblldb.so.13.0.0'])

SolarAquarion:
@yuyichao i decided to do also the lib32-llvm-git package stuff

yuyichao:
> lldb-git E: Dependency libedit detected and not included (libraries ['usr/lib/libedit.so.0'] needed in files ['usr/lib/liblldb.so.13.0.0'])
> lldb-git E: Dependency libxml2 detected and not included (libraries ['usr/lib/libxml2.so.2'] needed in files ['usr/lib/liblldb.so.13.0.0'])

Yes, I know exactly what you are talking about, but if you do want to do this, ignore namcap and scan for the dependencies yourself. The output from namcap is inconsistent and basically useless.......

archlinuxcn/lib32-llvm-git/PKGBUILD
```diff
@@ -12,7 +12,7 @@ url="https://llvm.org/"
 license=('custom:Apache 2.0 with LLVM Exception')
 makedepends=('git' 'cmake' 'ninja' 'python' 'python2'
              'lib32-gcc-libs' 'lib32-libffi' 'lib32-libunwind'
-             'lib32-libxml2' 'lib32-zlib')
+             'lib32-libxml2' 'lib32-zlib' 'gcc-multilib' 'llvm-git')
```
yuyichao, line 15: No `llvm-git` is not a build dependency. And gcc-multilib should not be included.

archlinuxcn/llvm-git/PKGBUILD
```diff
@@ -121,8 +121,8 @@ package_lldb-git() {
 
     pkgdesc="Next generation, high-performance debugger (git version)"
     url="https://lldb.llvm.org/"
-    depends=("clang-git=$pkgver-$pkgrel" "llvm-libs-git=$pkgver-$pkgrel")
-    optdepends=('llvm-git: for the man file')
+    depends=("clang-git=$pkgver-$pkgrel" "llvm-libs-git=$pkgver-$pkgrel" "python" "libedit" "libxml2")
```
yuyichao, line 124: If this is to include direct dependencies, I don't think this is the only case.

archlinuxcn/llvm-git/PKGBUILD
yuyichao, line 158: Same with the inconsistency here.

yuyichao (changes requested):

archlinuxcn/llvm-git/PKGBUILD
yuyichao, line 175: And this one is still wrong...

SolarAquarion:
@yuyichao so i should add depends on llvm-libs-git to all the depends?

yuyichao:
No, but if you want to be consistent, you should add libraries that are directly linked as dependencies.

archlinuxcn/llvm-git/PKGBUILD
SolarAquarion: Did I solve the inconsistency?

archlinuxcn/llvm-git/PKGBUILD
```diff
@@ -171,7 +171,7 @@ package_polly-git() {
 package_compiler-rt-git() {
     pkgdesc="Compiler runtime libraries for clang (git version)"
     url="https://compiler-rt.llvm.org/"
-    depends=("llvm-git=$pkgver-$pkgrel" 'gcc-libs')
+    depends=("llvm-git=$pkgver-$pkgrel" 'gcc-libs' "llvm-libs-git=$pkgver-$pkgrel")
```
yuyichao, line 174: `compiler-rt` should not depend on any of the LLVM packages actually.

SolarAquarion:
did i do the final fix?

yuyichao:
It's OK for `gcc-libs` now (at least you aren't adding any inconsistency here).

The other issue regarding directly linked dependency like `libedit` and `libxml2` are still not fixed.

yuyichao:
Current version looks OK. I'll squash and merge now before anything funny shows up on this branch again.

For future references, you probably want to use a different branch to open pull request instead of master.

SolarAquarion:
true enough, thanks

PR #2107, opened 2021-02-24
add mozilla-firefox-sync-server package

hamkido:


DDoSolitary:
Why not just use the package on AUR: https://aur.archlinux.org/packages/mozilla-firefox-sync-server/?

hamkido:
> Why not just use the package on AUR: https://aur.archlinux.org/packages/mozilla-firefox-sync-server/?
The compilation dependency on aur is wrong, you will see it when you use makechrootpkg to compile locally.
I only added mysql to make dependencies.
I will email the maintainer on aur about this dependency as soon as possible.

DDoSolitary:
Just checked it out and I can confirm that it lacks myql dependency. However, I would suggest patching the existing package using lilac.py rather than submitting a new package yourself.

hamkido:
> Just checked it out and I can confirm that it lacks myql dependency. However, I would suggest patching the existing package using lilac.py rather than submitting a new package yourself.


ok👌

PR #2116, opened 2021-02-26
Fix lilac.yaml

SolarAquarion:


SolarAquarion:
@yuyichao i hope you can squash and fix this.  Because i have something but like whatever

archlinuxcn/auracle-git/lilac.yaml
```diff
@@ -1,4 +1,8 @@
 maintainers:
   - github: colinkeenan
+  - github: solaraquarion 
 
-managed: false
+ build_prefix: extra-x86_64
```
yuyichao, line 5: Is yaml sensitive to leading space like this?

archlinuxcn/auracle-git/lilac.yaml
```diff
@@ -1,4 +1,8 @@
...
+ 
+ 
+update_on: vcs
```
yuyichao, line 8: And I think this is old syntax?

yuyichao (commented):
Commit structure looks good currently, not sure if there's anything random before the last force push. The commit and PR title should be slightly more descriptive. Not that important here but you should at least mention turn on auto build or sth like that...

Since this isn't my package I'll at least wait a few days before merging...

SolarAquarion:
It is

SolarAquarion:
@yuyichao yo

imlonghao:
Invalid YAML format.

https://github.com/archlinuxcn/repo/pull/2116/checks?check_run_id=1989237698

SolarAquarion:
fixed it i think

yuyichao:
> @yuyichao yo

Yes, and I was waiting for the fix.
Seems that the file has end-of-line whitespace and the indentation is a little confusing but it seems that it is at least valid now.

PR #2118, opened 2021-03-01
feat: optimus-manager@1.4-1 with fixes

outloudvi:
optimus-manager@1.4-1 is up on AUR, but the stock PKGBUILD is not working. This change updated the PKGBUILD as well as the patch in `lilac.py`.

BTW, buildlog: <https://build.archlinuxcn.org/packages/#/optimus-manager/logs/1614492157>


DDoSolitary:
I'll wait for the AUR maintainer to fix this.

PR #2138, opened 2021-03-16
addpkg: nordic-theme

hamkido:


SilverRainZ (approved):

archlinuxcn/nordic-theme/lilac.yaml
```diff
@@ -0,0 +1,13 @@
+maintainers:
+  - github: hamkido
```
imlonghao, line 2: The user has no public email.

imlonghao (changes requested):

archlinuxcn/nordic-theme/lilac.yaml
hamkido: I am sorry for my mistake.

> The user has no public email.

I have changed the mailbox to be publicly visible on github.

2021-03-17-112612_241x29_scrot

PR #2160, opened 2021-03-27
addpkg: rls-nightly

CoelacanthusHex:


CoelacanthusHex:
This package is added to rust-nightly package base, so I think it needs to be reviewed by its maintainer (@lilydjwg ).

PR #2193, opened 2021-04-18
mysql: use alias for icu

DuckSoft:


SilverRainZ:
No longer valid, closing.

PR #2199, opened 2021-04-19
zsh-syntax-highlighting-git: update AUR package after build

DDoSolitary:
The AUR package is unfortunatedly orphaned, so I adopted that package and I think it's a good idea to update it with lilac.


DDoSolitary:
The PKGBUILD changes are from an AUR user, who adopted the package, pushed a commit (which mainly syncs it with community/zsh-syntax-highlighting), and orphaned it shortly after.

PR #2202, opened 2021-04-21
Adding Mesa, automodapi, and doing some updates to llvm-git thanks to…

SolarAquarion:
… it for lilac to do it's job

Adding Mesa, automodapi, and doing some updates to llvm-git thanks to it for lilac to do it's job


SolarAquarion:
@yuyichao some updates because of one reason or another, like lordheavy did an update of his packages

yuyichao:
the changes to the LLVM packages makes sense and I assume they are correct.

SolarAquarion:
@yuyichao so?

PR #2258, opened 2021-05-14
fixing manfiles also adding lua to lldb

SolarAquarion:
also added lua53 to lldb


SolarAquarion:
@yuyichao this fixed llvm-git building because the man directories changed a bit in llvm

PR #2292, opened 2021-05-25
Update genie

rayfalling:


archlinuxcn/genie-systemd/lilac.yaml
```diff
@@ -4,12 +4,11 @@ maintainers:
 
 build_prefix: archlinuxcn-x86_64
 
+pre_build_script: aur_pre_build(maintainers=['rayfalling'])
```
DuckSoft, line 7: 404 - https://aur.archlinux.org/packages/genie-systemd

PR #2311, opened 2021-06-01
Add NODE_ENV to ferdi-git startup script.

ryanpetris:
As of [this commit](https://github.com/getferdi/ferdi/commit/38de1e0cf41b9d5527d405952e2d66e983ca2a5c) it appears that ferdi requires the NODE_ENV environment variable now otherwise it will start up in dev mode. The AUR package for ferdi-git also has has NODE_ENV set for a long time so this should probably be added here.

### DIFF
--- a/archlinuxcn/ferdi-git/ferdi.sh
+++ b/archlinuxcn/ferdi-git/ferdi.sh
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-exec electron '/usr/lib/ferdi/app.asar' "$@"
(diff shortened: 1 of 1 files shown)


yan12125:
Thanks for the patch! The package maintainer pushed a fix 5cf6f1398b0e097a9242fc228bbbd20ebeb2a333 (independently?).

ryanpetris:
Awesome, thank you. Yeah they must have found out the same way that I
did when I updated and it started in dev mode.

PR #2348, opened 2021-06-30
Fix missing header in anbox-git

Y7n05h:


yan12125:
Thanks for working on this. The missing header is added as part of a recent update [1]. However, anbox-git still fails for other reasons. You may want to check logs from [2].

[1] https://github.com/archlinuxcn/repo/commit/2bc794db0c8fa2ce24df53464228dcc20c915bc9#diff-412375d17bc9f49d37491103c633d3fd40d48b63a010adad58e0428cdb1ec83dR56
[2] https://build.archlinuxcn.org/packages/#/anbox-git

PR #2352, opened 2021-07-01
[Lib32-mesa-git] hopefully fixing the build

SolarAquarion:


archlinuxcn/lib32-mesa-git/lilac.yaml
```diff
@@ -5,9 +5,11 @@ build_prefix: multilib
 pre_build: vcs_update
 post_build: git_pkgbuild_commit
 repo_depends:
+    - llvm-git: llvm-git
```
yuyichao, line 8: I assume you’ll also need llvm libs if this is a build dependency problem.

PR #2369, opened 2021-07-19
fix(telegram-desktop-git): Fix missing jemalloc runtime depends

KunoiSayami:
After upgrade telegram-desktop-git, it crash after start.
```
$ telegram-desktop 
telegram-desktop: error while loading shared libraries: libjemalloc.so.2: cannot open shared object file: No such file or directory
```

It can working properly after install [extra/jemalloc](https://archlinux.org/packages/extra/x86_64/jemalloc/)


farseerfc (changes requested):
LGTM

PR #2536, opened 2021-11-14
change the update_on

starsareintherose:


SilverRainZ:
Please explian why you make this changes.

starsareintherose:
Update_on from aur-depends to the url-depends, making it update faster.

farseerfc (changes requested):
changing update_on alone is not enough to update the pkgver in PKGBUILD to follow the latest releases, we will need to `edit_file('PKGBUILD')` from lilac.py and use custom pre/post_build functions in order to be faster.
We can accept this change if you also fix lilac.py

PR #2610, opened 2021-12-29
fix the PKGBUILD of lilac-git

starsareintherose:


starsareintherose:
@lilydjwg

yan12125:
Thanks for keeping this package fresh and sorry for messing up the package on AUR. I'm not sure if syncing from this repo to AUR via update_aur_repo() is really a good idea. How about syncing from AUR to this repo via aur_pre_build/aur_post_build instead?

starsareintherose:
That's fine

lilydjwg:
python-sqlalchemy is now optional.

starsareintherose:
okay

yan12125:
Thanks for the update! Could you also add an entry to `update_on` to trigger updates in this repo when PKGBUILD on AUR is updated?

> python-sqlalchemy: sql

May be better as: `python-sqlalchemy: for recording build results in a database`

starsareintherose:
I've updated

yan12125:
Ah, forgot to mention, old `pre_build_script` and `post_build_script` blocks should be removed.

oldherl:
> Thanks for keeping this package fresh and sorry for messing up the package on AUR. I'm not sure if syncing from this repo to AUR via update_aur_repo() is really a good idea. How about syncing from AUR to this repo via aur_pre_build/aur_post_build instead?

I thought syncing from archlinuxcn to AUR was preferred over the other way around.

starsareintherose:
> Ah, forgot to mention, old `pre_build_script` and `post_build_script` blocks should be removed.

deleted

starsareintherose:
> > Thanks for keeping this package fresh and sorry for messing up the package on AUR. I'm not sure if syncing from this repo to AUR via update_aur_repo() is really a good idea. How about syncing from AUR to this repo via aur_pre_build/aur_post_build instead?
> 
> I thought syncing from archlinuxcn to AUR was preferred over the other way around.

both are okay, I am the maintainer of lilac-git in AUR, usually, I hope it can be auto-updated, I am lazy enough, so don't worry about this

yan12125:
> I thought syncing from archlinuxcn to AUR was preferred over the other way around.

Well, that depends case-by-case. For this package, syncing from [archlinuxcn] to AUR means that to update this package, @starsareintherose needs to submit a pull request, waiting for a few days for reviews and the lilac schedule. Such a delay can be avoided. I will merge this to unbreak the package for now. If there are better suggestions, feel free to propose!

PR #2629, opened 2022-01-18
fix: remove `npm --user root`

ouuan:
Since npm 7, the `--user` option is no longer available. `--user root` not only doesn't fix the permission problem, but also installs the package called "root", which conflicts with other packages that also have the `--user root` option.


imlonghao (changes requested):
`joplin` is following AUR package and your change will not effect, meanwhile, it's broken at this time. (#2611)

ouuan:
> [SilverRainZ](/SilverRainZ) added the [ upstream issue ](https://github.com/archlinuxcn/repo/issues?q=is%3Aopen+label%3A%22upstream+issue%22+sort%3Aupdated-desc) label

We can ignore joplin in this PR.

SilverRainZ:
> We can ignore joplin in this PR.

Okay, I removed the label.

SilverRainZ:
@im

imlonghao (approved):
LGTM

PR #2656, opened 2022-02-04
Change target to lowercase

JuanVCS:
So it actually applies over the generated binary


heavysink:
They changed to lower case again? Strange...

PR #2658, opened 2022-02-06
telegram-desktop-megumifox: allow reading font family name from env

oldherl:
allows the user to specify custom UI font family name by two environment variables:
TELEGRAM_DESKTOP_UI_FONT and TELEGRAM_DESKTOP_UI_FONT_MONO


CoelacanthusHex:
I think it is better to fallback to megunifox's implement when environment variable is empty.

CoelacanthusHex:
Oh, I'm blind

CoelacanthusHex:
In fact, since Qt's standard fonts are already used, Qt uses fontconfig. So you can write a fontconfig configuration to set the font of telegram, such as this.
https://github.com/CoelacanthusHex/dotfiles/blob/master/fontconfig/.config/fontconfig/conf.d/52-telegram-font.conf

PR #2669, opened 2022-02-15
addpkg: rust-analyzer-nightly

Y7n05h:


lilydjwg:
`source` needs to be updated too.

Y7n05h:
> `source` needs to be updated too.

`source` does not need to be changed, the required file already exists in the [first line](https://github.com/archlinuxcn/repo/pull/2669/files#diff-d3af0bff96089806b14f4ddb730800a6e85d4b517bb12fc8decc3b18b78f3790L21) of `source`.

lilydjwg:
ok.

Issue #2744, opened 2022-04-22
jdownloader2 文件不区分大小写重复


holymonson:
### 问题类型 / Type of issues

* 其它 / other

### 受影响的软件包 / Affected packages

* jdownloader2
* 整个`archlinuxcn`仓库

----

`jdownloader2`文件夹下包含`jdownloader`和`JDownloader`两个文件，在不区分大小写的文件系统下，会导致整个`archlinuxcn`仓库都没法checkout。
`jdownloader`这个文件完全可以用symlink代替。

oldherl:
正常来说，linux不应使用不区分大小写的文件系统。

holymonson:
我看了下，PKGBUILD里根本就没有用`jdownloader`这个文件，用`ln -s`代替了，所以就没必要留着这个文件了？

yan12125:
> 正常来说，linux不应使用不区分大小写的文件系统。

Yes, but this issue is still valid as the repository may be cloned on case-insensitive filesystems. See https://github.com/archlinuxcn/repo/pull/1096 and https://aur.archlinux.org/packages/jdownloader2#comment-853628.

yan12125:
Seems the current lilac does not allow removing a file downloaded from AUR. If I run rm after `aur_pre_build()`:

```yaml
pre_build_script: |
  aur_pre_build(maintainers=['Derson5'])
  run_cmd(['rm', '-f', 'jdownloader'])

post_build: aur_post_build
```

`aur_post_build()` fails as it cannot find the file `jdownloader`.

```
[D 04-22 20:34:35.815 cmd:88] running ['git', 'add', '-f', '--', 'JDownloader', 'JDownloaderHeadless', 'JDownloaderHeadlessCleanLogin', 'JDownloaderHeadlessCtl', 'PKGBUILD', 'functions.sh', 'jd-container16.png', 'jd-container22.png', 'jd-container24.png', 'jd-container256.png', 'jd-container32.png', 'jd-container48.png', 'jd-containers.desktop', 'jdownloader', 'jdownloader.desktop', 'jdownloader.install', 'jdownloader.service', 'jdownloader.xml', 'jdownloader16.png', 'jdownloader22.png', 'jdownloader24.png', 'jdownloader256.png', 'jdownloader32.png', 'jdownloader48.png'], not using pty, showing output
fatal: pathspec 'jdownloader' did not match any files
[...]

PR #2749, opened 2022-04-23
jdownloader2: remove an unused file and update AUR maintainer

yan12125:
This is an improved version for #1096. I forgot that this package syncs
from AUR.

Also removes lilac.py, which overrides lilac.yaml

Fixes https://github.com/archlinuxcn/repo/issues/2744


heavysink:
搞定了。谢谢！

PR #2792, opened 2022-05-29
(fixing and features) adding dependencies and fixing the build, but a…

SolarAquarion:
…lso removing some non necessary lines


yan12125:
This needs rebase. It's better to keep original pkgver, otherwise there will be frequent merge conflicts.

PR #2820, opened 2022-06-22
addpkg: watchman

pickfire:
> All releases since 4.9.0 have been internal Facebook releases, and Facebook is neither testing that the code will build on non-FB systems, nor willing to support people outside of Facebook. Everyone outside of Facebook are supposed to use the binary builds. Sadly, it's the way it is at Facebook open source.
>
> https://aur.archlinux.org/packages/watchman#comment-794701

不知道 watchman-bin 会不会比较好呢，facebook 已经好久没更新了


pickfire:
还是别了

PR #2870, opened 2022-08-07
fix: (llvm-git, lib32-llvm-git) the reasons why they can't install 

SolarAquarion:
I'm also adding more files in clang-git/usr/bin fixing conflicts in pacman

Next, i'm going to push mesa fixes/updates

### DIFF
--- a/archlinuxcn/llvm-git/PKGBUILD
+++ b/archlinuxcn/llvm-git/PKGBUILD
@@ -173,7 +173,7 @@ package_compiler-rt-git() {
 
     local _llvmver=$(echo "$pkgver" | grep -Po "(^[\.\d]+)")
 
-    _fakeinstall fakeinstall/usr/lib/clang/"$_llvmver"/lib/linux
+    _fakeinstall fakeinstall/usr/lib/clang/"$_llvmver"/lib/"$CHOST"
     _fakeinstall fakeinstall/usr/lib/clang/"$_llvmver"/include/{sanitizer,xray}
     _fakeinstall fakeinstall/usr/lib/clang/"$_llvmver"/share
 
@@ -195,7 +195,7 @@ package_clang-git() {
 
     _fakeinstall fakeinstall/usr/bin/*clang*
     _fakeinstall fakeinstall/usr/bin/{c-index-test,diagtool,find-all-symbols}
-    _fakeinstall fakeinstall/usr/bin/{hmaptool,modularize,scan-build,scan-view}
+    _fakeinstall fakeinstall/usr/bin/{hmaptool,modularize,scan-build,scan-build-py,scan-view,analyze-build,intercept-build}
     _fakeinstall fakeinstall/usr/bin/pp-trace
 
     _fakeinstall fakeinstall/usr/include/clang*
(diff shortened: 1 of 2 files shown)


archlinuxcn/lib32-llvm-git/PKGBUILD
```diff
@@ -4,7 +4,7 @@
 pkgbase=lib32-llvm-git
 pkgname=('lib32-llvm-libs-git' 'lib32-clang-git' 'lib32-llvm-git')
 pkgdesc='Low Level Virtual Machine (git version)'
-pkgver=16.0.0_r432160.786b503f66b1a
+pkgver=15.0.0_r425063.0bd645d3
```
SilverRainZ, line 7: This pkgver is older than the previous one, is this expected? Or you should bump the epoch.

SolarAquarion:
No, that was a mistake.  The runner should update the pkgver

SolarAquarion:
@SilverRainZ lilac run should bump the pkgver

archlinuxcn/mesa-git/PKGBUILD
```diff
@@ -91,7 +91,7 @@ package_vulkan-mesa-layers-git() {
   provides=('vulkan-mesa-layers')
   depends=('wayland' 'libxcb' 'python')
   conflicts=('vulkan-mesa-layer-git' 'vulkan-mesa-layers')
-  replaces=('vulkan-mesa-layer-git')
+  replaces=('vulkan-mesa-layer')
```
yuyichao, line 94: You must not replace official package and VCS package must also not replace non-VCS packages in general.

yuyichao (changes requested):

archlinuxcn/mesa-git/PKGBUILD
```diff
@@ -91,7 +91,6 @@ package_vulkan-mesa-layers-git() {
   provides=('vulkan-mesa-layers')
   depends=('wayland' 'libxcb' 'python')
   conflicts=('vulkan-mesa-layer-git' 'vulkan-mesa-layers')
-  replaces=('vulkan-mesa-layer-git')
```
yuyichao, line 94: Assuming he package used to be called vulkan-mesa-layer (without s) this should stay as is.

archlinuxcn/lib32-llvm-git/PKGBUILD
yuyichao: There’s no other change to this file? Then this is only going to cause conflicts….

archlinuxcn/llvm-git/PKGBUILD
```diff
@@ -173,7 +173,7 @@ package_compiler-rt-git() {
 
     local _llvmver=$(echo "$pkgver" | grep -Po "(^[\.\d]+)")
 
-    _fakeinstall fakeinstall/usr/lib/clang/"$_llvmver"/lib/linux
```
yuyichao, line 176: Why the change of path? The none git version seems to be installing to linux like the old version rather than CHOST.

yuyichao (changes requested):

archlinuxcn/llvm-git/PKGBUILD
SolarAquarion: it on my system and other build hosts was saying that there was no such thing as blah blah blah, and i realized that it changed it's logic to $CHOST

yuyichao (approved):

yuyichao:
Please avoid changing the pkgvers in PR since they can easily create merge conflicts.

Also, please don't include merge commits in PR since they are more difficult to be rebased out especially when they are there due to merge conflicts. I've rebased manually.

PR #2892, opened 2022-08-27
motrix: use nvm to build

zhullyb:
When we use nodejs16 to build it, it won't pass.
So we add 2 patches to fix the probelm.
However, a new issue occured that when we open motrix for a long time,
a pop-up windows will say "cb is not a function."
This time, we use nodejs14 build motrix to avoid it.


SilverRainZ (approved):
LGTM

PR #3002, opened 2022-11-03
fix: yuzu-early-access

sukanka:
`yuzu-early-access` fails to be built because `CMAKE_AR` is not defined.
For a detailed log, see https://build.archlinuxcn.org/packages/#/yuzu-early-access/logs/1667420698

Use `llvm-ar` since we are using `clang` (instead of `gcc`) as compiler.


heavysink:
f3cb5057dd60fbcf2179fd533bf83469c9bad057

PR #3027, opened 2022-11-11
rust-nightly: add optdepend for rust-analyzer-nightly

Y7n05h:


lilydjwg:
I prefer it to be `optdepends`, because the user may already have the source elsewhere (e.g. they pull the rustc git repo locally).

Y7n05h:
You are right. But `rust-src` is a dependency of `community/rust-analyzer`, do we need to be consistent with `community/rust-analyzer` ?

lilydjwg:
No. Being different than Arch's official packages is one of the repository's use cases.

Y7n05h:
OK, I have fixed it.

PR #3109, opened 2023-01-13
devtools-archlinuxcn: update to 20230105

yan12125:
[archlinuxcn commits](https://github.com/archlinuxcn/devtools/compare/archlinux:devtools:master...master) are rebased without any conflict. This time I test only the new archlinuxcn-specific patch (https://github.com/archlinuxcn/devtools/pull/1) and skip existing ones.

There is one noticeable change in [upstream commits](https://github.com/archlinux/devtools/compare/20221012...20230105): debug packages are enabled by default. There are already several debug packages in [archlinuxcn], so it should work in general. Possible issues may include broken builds with debug flags or disk usage for mirrors.

### DIFF
--- a/archlinuxcn/devtools-archlinuxcn/PKGBUILD
+++ b/archlinuxcn/devtools-archlinuxcn/PKGBUILD
@@ -2,11 +2,11 @@
 
 _pkgname=devtools
 pkgname=devtools-archlinuxcn
-pkgver=20221012
-pkgrel=4
-# curl https://api.github.com/repos/archlinuxcn/devtools/git/ref/tags/$pkgver-archlinuxcn2 | jq -r .object.sha
-_tag=1249cadb2edb5aa73477eb248dc854fcf365b479
-_upstream_pkgrel=2
+pkgver=20230105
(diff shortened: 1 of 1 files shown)


lilydjwg:
I'm a little concerned about debug by default. Some large packages (e.g. kernels and web browers) will potientially take much more time to generate separate debug symbols and compress the debug packages.

yan12125:
I assume few packages are affected. Using `options=(!debug)` for such packages and enabling debug packages by default for others is a good trade-off in my opinion.

By the way, I guess Linux kernels use kernel configs like `CONFIG_DEBUG_INFO` instead of environment variables.

lilydjwg:
We can try, and update any broken packages.

lilydjwg (approved):

yan12125:
Let's see ;)

yan12125:
The git tag hash was incorrect. I fixed it in f5e25c8e56d9e0a4a47dd601ce5bfcd563bdf7df.

https://build.archlinuxcn.org/packages/#/devtools-archlinuxcn

PR #3151, opened 2023-02-05
`miktex`: update_on alias: `boost-libs`

bryango:
Fix `error while loading shared libraries: libboost_locale.so.1.80.0: cannot open shared object file: No such file or directory`

### DIFF
--- a/archlinuxcn/miktex/lilac.yaml
+++ b/archlinuxcn/miktex/lilac.yaml
@@ -9,3 +9,4 @@ update_on:
     github: MiKTeX/miktex
     use_latest_release: true
   - alias: icu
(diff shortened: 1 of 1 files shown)


lilydjwg:
That alias is called `boost`: https://github.com/archlinuxcn/lilac/blob/master/lilac2/aliases.yaml

bryango:
> That alias is called `boost`: [archlinuxcn/lilac@`master`/lilac2/aliases.yaml](https://github.com/archlinuxcn/lilac/blob/master/lilac2/aliases.yaml?rgh-link-date=2023-02-05T04%3A32%3A48Z)

Thanks! Fixed accordingly.

PR #3158, opened 2023-02-08
Add nekoray

Molyuu:
Add nekoray,build from source.

But it can not pass CI by default without modification. Because `Go 1.19` is required while the latest version of `go` in community mirror is `1.20`. So I just do:
```bash
makedepends=(
... 'go=2:1.19.5-1' ...
)
```
I'm wondering if it is ok to do so.

### DIFF
--- a/archlinuxcn/nekoray/PKGBUILD
+++ b/archlinuxcn/nekoray/PKGBUILD
@@ -1,0 +1,72 @@
+# Maintainer: Molyuu <zhangjtroger@gmail.com>
+pkgname='nekoray'
+pkgver=2.14
+pkgrel=1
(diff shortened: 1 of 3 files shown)


hubutui:
Do not use `pkgname=epoch:pkgver-pkgrel`,  `epoch` & `pkgrel` are usually not used.

Since ArchLinux is a rolling distribution, do not depend on a specific pkgver, or it would not work when the deps update.

hubutui:
You could build it in a clean chroot, see also [wiki](https://wiki.archlinux.org/title/DeveloperWiki:Building_in_a_clean_chroot).

Molyuu:
Sorry for this stupid pr. I will learn more before creating pull request next time.

PR #3175, opened 2023-02-17
Update kicad-git

taotieren:


edward-p:
https://build.archlinuxcn.org/packages/#/kicad-git

Issue #3278, opened 2023-04-14
Build script of package 'gdb-multiarch' is not auto-updating


Donlon:
### 问题类型 / Type of issues

* 其它 / other

### 受影响的软件包 / Affected packages

* `gdb-multiarch`

----

In [lilac.yaml](https://github.com/archlinuxcn/repo/blob/master/archlinuxcn/gdb-multiarch/lilac.yaml#L12-L14) of package `gdb-multiarch`, the update source has been set to `aur` however the `PKGBUILD` file of it hasn't been updated since [12.1-2](https://aur.archlinux.org/cgit/aur.git/log/PKGBUILD?h=gdb-multiarch). Is this situation related to its orphaning (#3233)?

lilydjwg:
No, it's because [it failed to build](https://build.archlinuxcn.org/~imlonghao/#gdb-multiarch) (and the logs have been cleaned up).

Donlon:
So to investigate this problem, should I update the PKGBUILD file in the repo manually to trigger a rebuild and to get the build logs again? I have tested the script and it works on my system.

lilydjwg:
Yes.

PR #3279, opened 2023-04-14
gdb-multiarch: update build script to diagnose failed builds

Donlon:
See #3278


lilydjwg:
Please also update `lilac.yaml` to adopt the package so that error reports go to you.

PR #3305, opened 2023-05-07
gdb-multiarch: disable build for simulators

Donlon:
Currently the gdb's sim folder contains problematic code that cannot be compiled and linked without any error if LTO option is on. However the simulator executables are not added into the output package and have been probably obsoleted in many Linux distributions, like [Debian](https://salsa.debian.org/gdb-team/gdb/-/blob/debian/13.1-2/debian/rules?ref_type=tags#L204). To avoid failures when building the package with Archlinux's devtools, the '--disable-sim' option is added to the configuration command-line to disable these files. In addition, syncing with upstream AUR repository is temporarily disabled until the upstream no longer has this problem, but I'm not sure whether it's necessary to and how to add another sources, like the official release, to keep the package always up-to-date.

### DIFF
--- a/archlinuxcn/gdb-multiarch/PKGBUILD
+++ b/archlinuxcn/gdb-multiarch/PKGBUILD
@@ -2,7 +2,7 @@
 
 pkgname=gdb-multiarch
 pkgver=13.1
-pkgrel=1
+pkgrel=2
 pkgdesc='The GNU Debugger for all gdb supported architectures (i386/arm/mips...)'
 arch=(i686 x86_64)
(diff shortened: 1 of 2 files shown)


Donlon:
@lilydjwg Hi, it seems that there's no any person assigned to this PR. Could you please help with the updates?

lilydjwg:
You can track upstream but you need to write a corresponding `pre_build()` function to update `PKGBUILD`. You can also track AUR and edit the file in `pre_build()`.

PS: next time try to @ whoever merges your first pr for the package.

PR #3326, opened 2023-06-02
Update PKGBUILD

tourist-lby:
If you use `libdisplay-info`, you may encounter the following error when running `Hyprland`:
```
Hyprland:error while loading shared libraries: libdisplay-info.so.2: cannot open shared object file
```
Upon using `pkgfile`, it was discovered that the library `libdisplay-info.so.2` is only available in `libdisplay-info-git`.

Please note that `pkgfile` is a command-line utility used in some Linux distributions to search for package files related to a specific library or executable. It helps identify which package provides a particular file or library.

### DIFF
--- a/archlinuxcn/hyprland-nvidia/PKGBUILD
+++ b/archlinuxcn/hyprland-nvidia/PKGBUILD
@@ -38,7 +38,7 @@ depends=(
 	vulkan-validation-layers
 	xorg-xwayland
   libliftoff
-	libdisplay-info)
+	libdisplay-info-git)
 makedepends=(
 	git
 	cmake


hubutui:
@Integral-Tech
You add `libdisplay-info` as depends, but specify `libdisplay-info-git` in `lilac.yaml`, which make lilacbot build hyprland-nvidia with `libdisplay-info-git` and lead to this issue.
https://github.com/archlinuxcn/repo/blob/f3dd72d701330486accaa0d1d2166036a9057c37/archlinuxcn/hyprland-nvidia/PKGBUILD#L41

https://github.com/archlinuxcn/repo/blob/f3dd72d701330486accaa0d1d2166036a9057c37/archlinuxcn/hyprland-nvidia/lilac.yaml#L9

Integral-Tech:
Fixed in commit https://github.com/archlinuxcn/repo/commit/69780fa74eb97a77143365348e33081b8dbf8575.

PR #3366, opened 2023-07-09
kmonad: fix build error

ZeekoZhu:
Kmonad has replaced kmoand.service with kmonad@.service. This PR has updated the corresponding install script accordingly.


rocka:
You probably should add a [`post_upgrade` hook](https://wiki.archlinux.org/title/PKGBUILD#install), to inform user about this change.

ZeekoZhu:
@rocka The install script will display a message informing the user that the kmonad.service file has been replaced by kmonad@.service.

PR #3378, opened 2023-07-23
linux-bcachefs-git: new package

wszqkzqk:
Kent Overstreet's bcachefs filesystem


CoelacanthusHex:
We may no longer need this package because Linux 6.7 will become stable next week.

PR #3429, opened 2023-08-31
telegram-desktop-megumifox: bundle glibmm in as static lib, update to 4.9.3

chitang233:
merged from https://gitlab.archlinux.org/archlinux/packaging/packages/telegram-desktop/-/commit/5be98e37abe190f2b6bec5d692c0316c47b6f9ff


oldherl:
I would prefer stick to the current version

PR #3442, opened 2023-09-09
fix(mesa-git): Changes from mesa on upstream, adding more packages an…

SolarAquarion:
…d other stuff

it wasn't properly building before

fix pkgver

remove outdated build depends


SolarAquarion:
@lilydjwg

PR #3471, opened 2023-09-20
addpkg: yass-proxy

Chilledheart:


Chilledheart:
YASS is a small tool provide similar functionality like naiveproxy but rewritten from scratch. It has a GUI client (gtk4) and supports chinese.
See https://letshack.info/2023/09/16/YASS-Chinese-Translation/ and https://github.com/Chilledheart/yass/wiki.

PR #3481, opened 2023-09-29
devtools-archlinuxcn: update to 1.0.4

yan12125:
Also replaces bzr by breezy

See: https://gitlab.archlinux.org/archlinux/packaging/packages/devtools/-/commit/50f0b46b1c31f4e478e597688ce825ad68942340
See: https://archlinux.org/todo/change-dependency-from-bzr-to-breezy/

---

I only do basic smoke testing by building devtools-archlinuxcn itself.

There are no rebase conflicts, and [upstream commits](https://github.com/archlinux/devtools/compare/v1.0.3...v1.0.4) look trivial.

### DIFF
--- a/archlinuxcn/devtools-archlinuxcn/PKGBUILD
+++ b/archlinuxcn/devtools-archlinuxcn/PKGBUILD
@@ -4,10 +4,10 @@
 _pkgname=devtools
 pkgname=devtools-archlinuxcn
 epoch=1
-pkgver=1.0.3
-pkgrel=2
+pkgver=1.0.4
+pkgrel=1
 # curl https://api.github.com/repos/archlinuxcn/devtools/git/ref/tags/$pkgver-archlinuxcn1 | jq -r .object.sha
-_tag=0589cd09d493bd8c80f4e2c20a80dd93a38791d5
+_tag=ac4d056aa8254c8487ed93834a6d7e7b1e63e153
 _upstream_pkgrel=1
 pkgdesc='Tools for Arch Linux package maintainers, archlinuxcn fork'
 arch=('any')
@@ -29,7 +29,7 @@ depends=(
   sed
   util-linux
 
-  bzr
+  breezy
   git
   mercurial
   subversion


lilydjwg:
The [`--slice=` argument for nspawn](https://github.com/archlinux/devtools/compare/v1.0.3...v1.0.4#diff-20fd0a21d9260a1ec7d9bc6d627de479e7ece2bcbe49061810e42db54fc029dcL58) needs to be removed and `--keep-unit` be kept, otherwise lilac won't be able to track CPU and memory usage.

yan12125:
I'm not familiar with systemd slices, so I just reverted the relevant commit at https://github.com/archlinuxcn/devtools/commit/657fd731d2be5c8fae2b74c4d27d6e2025526c71. Please consider discussing requirements of Arch Linux CN community with upstream to figure out a robust approach for arch-nspawn. There are already too many archlinuxcn-specific patches, and I prefer to avoid more.

lilydjwg (approved):

lilydjwg:
I'd like to try [this change](https://github.com/archlinuxcn/devtools/commit/a1d0e111d36e6549983ab1b912bf082fc88b75a0).

lilydjwg:
> There are already too many archlinuxcn-specific patches, and I prefer to avoid more.

Last time the upstream response gave me the impression that they only want to support their own use cases. Has this been changed (or did I misunderstand)?

yan12125:
> I'd like to try [this change](https://github.com/archlinuxcn/devtools/commit/a1d0e111d36e6549983ab1b912bf082fc88b75a0).

Nice, looks good. Did you test it with lilac?

> Last time the upstream response gave me the impression that they only want to support their own use cases. Has this been changed (or did I misunderstand)?

I'm afraid that may still be the case from time to time. We need to face it sooner or later, anyway.

yan12125:
> No. I only checked the regex locally. I don't have another instance of lilac to test with....

No problem. I use systemd-run to simulate lilac behavior, and it seems working as intended.

Let's wait for a while to see if others have comments.

yuyichao (approved):
Looks good in areas other than the mounting options.

yan12125:
Thanks! Let's try it out.

PR #3505, opened 2023-10-20
screengrab-git: Rename kf5 packages

yan12125:
I maintain https://aur.archlinux.org/packages/screengrab-git but not this package, so syncing PKGBUILD is not trivial. This time, I update both manually.

### DIFF
--- a/archlinuxcn/screengrab-git/PKGBUILD
+++ b/archlinuxcn/screengrab-git/PKGBUILD
@@ -4,12 +4,12 @@
 _pkgname=screengrab
 pkgname=$_pkgname-git
 pkgver=2.6.0.15.g896ccd7
-pkgrel=1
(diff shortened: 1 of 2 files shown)


ykelvis:
maybe also change lilac.yaml so this will follow updates on aur? also i think you should consider adopting this package, it's better for same package have same maintainer.

yan12125:
> maybe also change lilac.yaml so this will follow updates on aur?

Thanks, that makes sense.

> also i think you should consider adopting this package, it's better for same package have same maintainer.

For this package it may not matter. There are only few PKGBUILD commits [1] in the past years, and I'd like to avoid the maintenance burden for a binary package.

[1] https://aur.archlinux.org/cgit/aur.git/log/?h=screengrab-git

PR #3552, opened 2023-12-06
feat(archlinuxcn-mirrorlist): update package

Zenvie:


archlinuxcn/archlinuxcn-mirrorlist/lilac.yaml
```diff
@@ -4,3 +4,4 @@ maintainers:
 update_on:
   - source: github
     github: archlinuxcn/mirrorlist-repo
+    path: archlinuxcn-mirrorlist
```
auxiliarypower, line 7: How can I write rules to scan only the files in the root directory and exclude the subdirectories?

PR #3611, opened 2024-01-22
fix(ttfautohint): fix errors reported by namcap

Zenvie:
https://github.com/alerque/aur/pull/72


Zenvie:
@Y7n05h Would it be possible for you to add me as a co-maintainer?

PR #3616, opened 2024-01-25
telegram-desktop-{lily,megumifox}: Rebuild on libvpx soname bump

q234rty:
`readelf -d /usr/bin/telegram-desktop` contains libvpx.so.8 so it's a direct dependency.

The PKGBUILD should also probably be modified to include libvpx as a dependecy...


Zenvie:
https://t.me/archlinuxcn_group/3050055

PR #3621, opened 2024-01-27
(fix): (spirv-headers-git) increase epoch and see if that fixes the errors

SolarAquarion:


lilydjwg:
Isn't that caused by [upstream tags](https://github.com/KhronosGroup/SPIRV-Headers/tags) and [AUR pkgver](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=spirv-headers-git#n21) don't match?

The previous fix was overridden by AUR because `aur_pre_build` pulls files from AUR and override things. Maybe you can just stop following AUR's PKGBUILD?

SolarAquarion:
Sounds good.

PR #3705, opened 2024-03-19
fix(llvm-git): Fix the build and pkgver

SolarAquarion:
it wasn't building and now it's building.  I also moved things to the source

### DIFF
--- a/archlinuxcn/llvm-git/PKGBUILD
+++ b/archlinuxcn/llvm-git/PKGBUILD
@@ -1,5 +1,5 @@
 #!/usr/bin/bash
-
+# Maintainer: Solomon Choina <shlomochoina@gmail.com>
 # Based on LordHeavy's LLVM
 pkgbase=llvm-git
 pkgname=('lldb-git' 'lld-git' 'polly-git' 'compiler-rt-git' 'clang-git' 'spirv-llvm-translator-git' 'llvm-libs-git' 'llvm-git')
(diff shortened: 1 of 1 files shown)


archlinuxcn/llvm-git/PKGBUILD
```diff
@@ -53,9 +58,7 @@ prepare() {
     # llvm-project contains a lot of stuff, remove parts that aren't used by this package
     rm -rf "$srcdir"/fakeinstall
 
-  cd llvm/projects
-  git clone https://github.com/KhronosGroup/SPIRV-LLVM-Translator.git
-
+  cp -r $srcdir/SPIRV-LLVM-Translator llvm/projects
 }
 
 build() {
```
Zenvie, line 64:
```
==> ERROR: A failure occurred in build().
```
https://fars.ee/Nclk.log

SolarAquarion: the issue is that there isn't enough storage on your device.

SolarAquarion: "/build/llvm-git/src/_build/bin/llvm-exegesis" to
  "/build/llvm-git/src/fakeinstall/usr/bin/llvm-exegesis": No space left on
  device.

Zenvie: I'm using a build server provided by the Arch Linux CN community, so insufficient storage space is unlikely.

SolarAquarion: so the size of the build server needs to be expanded

lilydjwg: @Zenvie you ran out the tmpfs space for the chroot by default. lilac has been instructed to `-r` to choose another place for it.

oldherl: @lilydjwg Is there an option in lilac.yaml that can instruct lilac to build the package outside tmpfs?

Zenvie: > @Zenvie you ran out the tmpfs space for the chroot by default. lilac has been instructed to use `-r` to choose another place for it.

This parameter has the unintended consequence of creating files that I lack the necessary permissions to delete.

SolarAquarion:
you couldn't delete those files before.  But I just moved it from a dirty
method to a clean method.  I don't think there should be a git clone within
git.  You would be having the same issue if it was a git submodule

archlinuxcn/llvm-git/PKGBUILD
lilydjwg: @oldherl it's in [lilac.py](https://github.com/archlinuxcn/lilac/blob/master/schema-docs/lilac-py-fields.md).

@Zenvie yes....

Issue #3748, opened 2024-04-01
64gram-desktop Cannot mix incompatible Qt library (6.6.2) with this library (6.6.3)


asukaminato0721:
### 问题类型 / Type of issues

* 打包错误 / packaging error

### 受影响的软件包 / Affected packages

*  64gram-desktop

----

Cannot mix incompatible Qt library (6.6.2) with this library (6.6.3)

asukaminato0721:
https://github.com/archlinuxcn/repo/issues/3744

PR #3749, opened 2024-04-01
https://github.com/archlinuxcn/repo/pull/3746/files

asukaminato0721:
#3748


lilydjwg:
It still fails: https://build.archlinuxcn.org/imlonghao-api/pkg/64gram-desktop/log/1711956477

libXdamage.so is missing.

PR #3794, opened 2024-04-29
librewolf: fix missing icon when running under wayland

Integral-Tech:
Currently, when running LibreWolf under Wayland, the icon will fallback to the Wayland default one.

Setting the correct ```MOZ_APP_REMOTINGNAME``` can fix the icon.


Integral-Tech:
@heavysink

Integral-Tech:
Fixed in the latest update from AUR.

Issue #3707, opened 2024-03-20
Use extra/llvm for mesa-git and maybe add a seperate package for mesa-git with llvm-git


q234rty:
### 问题类型 / Type of issues

* 其它 / other

### 受影响的软件包 / Affected packages

* mesa-git

----
Currently extra/llvm is [enough](https://gitlab.freedesktop.org/mesa/mesa/-/blob/62b490fa923d2dabe232e5bacd1924fdf254fd48/meson.build#L1760-1766) for mesa-git. Moreover, llvm-git breaks any package in the official repos that depend on `libLLVM-17.so`. If people really want mesa-git with llvm-git, it might be better to create a seperate package.

CC @SolarAquarion

lilacbot:
Some maintainers (perhaps outside contributors) cannot be assigned: @solaraquarion

oldherl:
these packages were removed

PR #3844, opened 2024-05-27
feat(mesa-git): a lot of updates

SolarAquarion:
it has bug fixes, it has packaging changes it has so many changes


q234rty:
Would be great if you can consider #3707.

PR #3928, opened 2024-07-04
clash-verge: fix mihomo name

Kiri2002:
The new version of clash-verge call verge-mihomo which is just clash-meta. And the process-name need to be verge-mihomo also, so here use a hard link


Kiri2002:
This should fix https://github.com/archlinuxcn/repo/issues/3923 and https://github.com/archlinuxcn/repo/issues/3924

PR #3929, opened 2024-07-05
clash-verge: soft link clash-meta to verge-mihomo

Kiri2002:
Sorry for the mistake, the hard link will point to the compile machine's clash-meta.
And the soft link also can create a new progress name which is verge-mihomo.
So here should be soft link


Kiri2002:
@Integral-Tech This time maybe will works...

PR #3943, opened 2024-07-15
calamares: add pkg and deps

Kiri2002:
Add new package calamares and it's dependency


Kiri2002:
- calamares
- ckbcomp
- kinitcpio-openswap

mostly copy from BioArchlinux.
Fixed the license:
GPL2 only for ckbcomp ("GPL2" before)

MIT for kinitcpio-openswap ("UNKNOWN" before, but found the MIT license just at the folder)

oldherl:
You are adding to the wrong place!

Kiri2002:
oh my god, so sorry for that!

PR #3945, opened 2024-07-15
calamares: add pkg and deps

Kiri2002:
same as https://github.com/archlinuxcn/repo/pull/3943 which put in wrong place.

https://github.com/archlinuxcn/repo/commit/40f30afdf2ae3d4802ca518fd3171d5201ed2e2a revert the bad commit.

Here put  it in right place


archlinuxcn/calamares/calamares-3.3.8-1-x86_64.pkg.tar.zst-namcap.log
oldherl, line 1: Don't add those garbage files.

Kiri2002: sorry for that mistake😱😱😱, deleted.

oldherl (approved):

oldherl:
@Kiri2002 Please keep in mind that you are using the "staging" feature. You would need to examine and move the pkg file each time it's generated.

PR #3951, opened 2024-07-18
v2raya: depends on v2raya-core

rocka:
xray provides `v2raya-core` since https://github.com/archlinuxcn/repo/commit/a45645ba7ea57e2d3b0753b5b41667f2e97df87b

### DIFF
--- a/archlinuxcn/v2raya-core/PKGBUILD
+++ b/archlinuxcn/v2raya-core/PKGBUILD
@@ -1,0 +1,14 @@
+# Maintainer: Rocket Aaron <i at rocka dot me>
+
+pkgname='v2raya-core'
+pkgdesc='v2ray core for v2rayA'
+pkgver=1
+pkgrel=1
+url='https://github.com/v2fly/v2ray-core'
+license=('MIT')
+depends=('v2ray')
+arch=('any')
+
+package() {
+    # nothing to package
+}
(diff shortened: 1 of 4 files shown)


archlinuxcn/v2raya/PKGBUILD
```diff
@@ -7,8 +7,8 @@ pkgrel=1
 pkgdesc="A web GUI client of Project V which supports VMess, VLESS, SS, SSR, Trojan and Pingtunnel protocols"
 arch=('x86_64')
 url="https://github.com/v2rayA/v2rayA"
-license=('AGPL3')
-depends=('glibc')
+license=('AGPL-3.0')
```
zjuyk, line 10: Thanks for the correction!

archlinuxcn/v2raya-core/PKGBUILD
```diff
@@ -0,0 +1,14 @@
...
+pkgver=1
+pkgrel=1
+url='https://github.com/v2fly/v2ray-core'
+license=('MIT')
+depends=('v2ray')
+arch=('any')
+
+package() {
+    # nothing to package
+}
```
lilydjwg, line 14: This is bash...

rocka: didn't know this before 🥲

should have wrote `:;` instead?

lilydjwg: Or `true`. `;` is not needed.

PR #3964, opened 2024-07-26
fix(sing-geosite): checksum failed

everyx:


rocka:
Checksum update should be handled by lilac, rather than manually. Also pkgver in this PR is outdated. Closing.

PR #4002, opened 2024-08-22
qadwaitadecorations-qt: change maintainer and update_on source

everyx:


oldherl:
You didn't fix anything in the PKGBUILD. It was failing before your change and has just failed again. Please fix it.

PR #4117, opened 2024-12-15
devtools-archlinuxcn: update to 1.3.0

yan12125:
* Update dependencies following https://gitlab.archlinux.org/archlinux/packaging/packages/devtools/-/commit/a58ac354b286129fac1fa01f82ce135dce422829

---

* One conflict between https://github.com/archlinuxcn/devtools/commit/1ddd4c93c471d9adeb892d03f3591299dd1568f1 and https://gitlab.archlinux.org/archlinux/devtools/-/merge_requests/287. I didn't test it as I'm not sure what's the rationale behind the new local commit.
* [Upstream commits](https://github.com/archlinux/devtools/compare/v1.2.1...v1.3.0) include some non-trivial changes and may need more checking.

### DIFF
--- a/archlinuxcn/devtools-archlinuxcn/PKGBUILD
+++ b/archlinuxcn/devtools-archlinuxcn/PKGBUILD
@@ -5,8 +5,8 @@
 _pkgname=devtools
 pkgname=devtools-archlinuxcn
 epoch=1
-pkgver=1.2.1
-pkgrel=2
+pkgver=1.3.0
+pkgrel=1
 _tag=$pkgver-archlinuxcn1
 _upstream_pkgrel=1
 pkgdesc='Tools for Arch Linux package maintainers, archlinuxcn fork'
@@ -24,7 +24,9 @@ depends=(
   expac
   fakeroot
   findutils
+  glow
   grep
+  gum
   jq
   openssh
   parallel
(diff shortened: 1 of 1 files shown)


lilydjwg:
I'll update that commit because nspawn name conflicts are still possible: https://build.archlinuxcn.org/imlonghao-api/pkg/python-scikit-image/log/1734154026

yan12125:
Is something like `--keep-unit --machine="arch-nspawn-$$"` a solution?

lilydjwg:
I think so. I'll give it a try tomorrow.

lilydjwg:
It seems to work. I've pushed 283b4b36a96f311c6c7c7787f2f1a31fd1738b47 for the change.

I still don't understand why the linked error occurred though, since the chroot is protected by a lock.

> Upstream commits include some non-trivial changes and may need more checking.

Most of them are about things lilac doesn't use, so I think they are fine.

yan12125:
> It seems to work. I've pushed 283b4b36a96f311c6c7c7787f2f1a31fd1738b47 for the change.

Thank you, I re-tagged and updated this pul request.

---

> I still don't understand why the linked error occurred though, since the chroot is protected by a lock.

Do you mean this?
```
Mount point '/run/systemd/nspawn/unix-export/extra-x86-64-root' exists already, refusing.
```
This seems not about the chroot itself, but a new feature in systemd 256, which allows sharing UNIX domain sockets among the host and nspawn-ed containers [1][2]. Seems there is no way to disable it [3].

[1] https://mastodon.social/@pid_eins/111725470117923127
[2] https://github.com/systemd/systemd/blob/v256/NEWS#L425
[3] https://github.com/systemd/systemd/blob/v257/src/nspawn/nspawn.c

lilydjwg (approved):

yan12125:
Maybe it's caused by a race condition from flock + unlink combination:
https://github.com/archlinuxcn/lilac/blob/92d9e4a43d0cce5184ce96f344842f601590a097/scripts/build-cleaner#L37
https://www.dingmos.com/index.php/archives/142/

lilydjwg:
I meant [the lock in `archbuild`](https://gitlab.archlinux.org/archlinux/devtools/-/blob/master/src/archbuild.in?ref_type=heads#L104).

`build-cleaner` skips the `root` directory so it's not it. Also the race condition is harmless here: since we delete the file just before we release the lock, nothing is left to do with the lock. The only possible issue is that two instances try to delete the file at the same time but that's still OK.

yan12125:
> build-cleaner skips the root directory so it's not it.

Didn't notice that. No idea then.

> The error occurred only once, so probably we can ignore it for now.

okay

yan12125:
Looks like yuyichao is busy these days. Let's just move on.

PR #4123, opened 2024-12-22
ecapture: force rebuild

yingziwu:
ecapture: force rebuild


lilydjwg:
You can use [triggerabuild](https://build.archlinuxcn.org/triggerabuild/) to rebuild packages.

PR #4198, opened 2025-02-23
mpv-git: move ffnvcodec-headers and wayland-protocols to makedepends

yejingchen:


CoelacanthusHex:
It's similar to glib-devel. It's the dependency when it was a makedepends, not makedepends. Pacman can't differ depends and the dependency when it was a makedepends, so it must be depends now. Unless we use a split package named mpv-devel-git.

yejingchen:
Ok, let's keep it that way.

Issue #4135, opened 2024-12-25
 Python 3.13 Rebuild


lilydjwg:
### 问题类型 / Type of issues

* 其它 / other

### 受影响的软件包 / Affected packages

* [x] asahi-fwextract (this is an "any" package for aarch64 and should automatically update when the alarm python version is updated)
* [x] python-sip4 (@Integral-Tech)
* [x] python-pulsectl (@axionl)
* [x] opendrop (@heavysink)
* [x] python-customtkinter (@heavysink)
* [x] cambalache (@kiri2002)
* [x] bypy-git (@petronny)
* [x] papis (@roaldclark)
* [x] python-cloudscraper (@roaldclark)
* [x] python-ufo2ft (@roaldclark)
* [x] python-soco (@tttturtle-russ)
* [x] debspawn (@weilinfox)
* [x] ruyi (@weilinfox)
* [x] python-xingque (@weilinfox)

----

List updated on 2024-12-31 10:00.

These packages contain files for python 3.12, and should be updated to python 3.13 now. It's a similar issue as #3786. This list is generated by
`pacman -Fq "/usr/lib/python3.12/site-packages/" | awk -F/ '$1 == "archlinuxcn" { print "* [ ] " $2 }' | PYTHONPATH=./lilac/ ./lilac/scripts/at-maintainer | sort -k5` where `./lilac` is a clone of [lilac](https://github.com/archlinuxcn/lilac) and `./who_depends_this_lib /data/repo/x86_64/archlinuxcn.files.tar.gz 'libpython3\.12' --dep-pkgname python` on build machine where `who_depends_this_lib` comes from [misc_scripts](https://github.com/archlinuxcn/misc_scripts).

Most packages have been rebuilt by lilac, but the above are missing.

For packages failed with its dependencies not rebuilt successfully, they will be automatically rebuilt on lilac's next run once the dependencies are fixed.

For other packages, please investigate.

Please only tick the checkbox only AFTER the package is actually BUILT successfully.

After updating the files, organization members can use [triggerabuild](https://build.archlinuxcn.org/triggerabuild/) to build packages as soon as lilac is idle.

lilacbot:
Some maintainers (perhaps outside contributors) cannot be assigned: @yuyichao, @sihuan, @weilinfox, @tttturtle-russ, @kiri2002, @donlon, @ykelvis

oldherl:
I just added the missing update_on "alias: python" for these packages and they have been successfully built.
[...]

PR #4216, opened 2025-03-08
bypy-git: fix python3.13

oldherl:
Not updating from AUR because AUR is still on python3.12


oldherl:
A part of #4135

oldherl:
Not needed anymore.

PR #4246, opened 2025-04-04
citron-git: Add package

sukanka:
This PKGBUILD is adapted from `yuzu-early-access`. I also clean some unnecessary sources, such as 
- libadrenotools: only used by android
- breakpad: only for windows
- tzdb_to_nx: with the zip file, it's not needed now.

This PKGBUILD is tested by `archlinuxcn-x86_64-build`


archlinuxcn/citron-git/PKGBUILD
terinasia, line 1: test comment

PR #4324, opened 2025-06-07
Update PKGBUILD, sync with AUR

davidbepo:
pkg fails to build otherwise


davidbepo:
synced again

davidbepo:
@poscat0x04

PR #4364, opened 2025-07-12
fix(fcitx5-pinyin-moegirl): use fixed file name

DeepChirp:
Refer to the [`rime-pinyin-zhwiki`](https://archlinux.org/packages/extra/any/rime-pinyin-zhwiki/) package on the official extra repository, we should use a fixed file name. If not, users will need to update the `rime` or `fcitx5` configuration file frequently, causing trouble.


imlonghao:
make sense, fix at f8ce5ba9d83ca011bbc14ce55fe3b2b07260cf9d

DeepChirp:
@imlonghao Perhaps we should modify pkgrel to allow the corrected packages to be distributed?

imlonghao:
> @imlonghao Perhaps we should modify pkgrel to allow the corrected packages to be distributed?

done. https://github.com/archlinuxcn/repo/commit/ae50901fd437b53dd181c34ff8bcc9d9b7399a5f

PR #4401, opened 2025-08-09
Merge patch to fix missing icon and package conflict

CoreanXka:
Sync patch with AUR comments

### DIFF
--- a/archlinuxcn/zen-browser-portable/PKGBUILD
+++ b/archlinuxcn/zen-browser-portable/PKGBUILD
@@ -15,13 +15,9 @@ depends=(
 	"portable"
 )
 
-optdepends=(
-	"zen-browser-i18n-zh-cn: Language pack for Zen Browser (zh-CN)"
-	"zen-browser-i18n-zh-tw: Language pack for Zen Browser (zh-TW)"
-	"zen-browser-i18n-ja: Language pack for Zen Browser (ja)"
-)
+optdepends=()
 
-provides=("zen-browser=${pkgver}")
+provides=(zen-browser=${pkgver})
 conflicts=(zen-browser)
 
 makedepends+=(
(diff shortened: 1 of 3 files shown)


Kimiblock:
In summary, please revert all modifications to the portable config, desktop files and start script.

Kimiblock:
I'm wearing portable maintainer hat on this comment:

In the future, we'd like to enforce using application IDs for /usr/lib/portable/info. That means portable will be unable to find the configuration if this change is merged. The reason behind this is to make a unified config store that supports overriding permissions, etc.

CoreanXka:
> I'm wearing portable maintainer hat on this comment:
> 
> In the future, we'd like to enforce using application IDs for /usr/lib/portable/info. That means portable will be unable to find the configuration if this change is merged. The reason behind this is to make a unified config store that supports overriding permissions, etc.

The actual WMClass of zen-browser at runtime is “zen-browser”. Testing has shown that only when WMClass is set to “zen-browser” and the appID is also “zen-browser” can KDE fully match the icon and display it correctly. In the portable-config, the appID is set to “org.mozilla.zen”. According to the note, this value must not conflict with the actual appID. Testing has confirmed that it can be any value other than “zen-browser”, so I have not made any changes to it. Currently, zen-browser is working well on my computer (except that the portable version creates an “org.mozilla.zen.desktop” file when zen is running, which is automatically deleted after the program is closed). What you mean is that the portable version will change its appID handling logic in the near future, right?

Kimiblock:
The StartupWMClass is a bug, yes.

Regarding .desktop files. This is considered a KDE issue, it should not prefer application provided app_id over sandbox provided one. GNOME does correctly identify all sandboxed application's window. Additionally, the desktop file must be in reverse DNS for XDP to work.

If you want it to work under KDE, use the stub desktop file in the PKGBUILD that is used for MPRIS.

Kimiblock:
The _portableConfig environment only affects the way portable finds the app config, it does not affect the actual application ID.

Kimiblock:
> According to the note, this value must not conflict with the actual appID

No, this simply requires the appID to be unique in the scope of portable applications.

Kimiblock:
> except that the portable version creates an “org.mozilla.zen.desktop” file when zen is running

This is because missing .desktop file will cause XDP to crash previously

Kimiblock:
> What you mean is that the portable version will change its appID handling logic in the near future, right?

Yes and no. The part that will change is the logic of finding portable config files, which will be forced to tie to appID to reduce confusion

CoreanXka:
The `_portableConfig` variable affects how portable locates its configuration, with the path being `/usr/lib/portable/info/${_portableConfig}/config`. However, differences in this path can influence whether KDE can find the icon during Alt+Tab switching. When `_portableConfig` is set to `zen-browser`, the icon appears correctly, which is why I changed its value. Do you know the reason behind this behavior?
However, could you trouble to test this PKGBUILD under GNOME. If it works fine, at least this change would temporarily resolve the icon issue under KDE.
Also, if you are certain that this PR cannot be merged, I can revert this changes, and remove the `optdepends` content from my previous commit, as it seems to cause a mild circular dependency issue.

Kimiblock:
Have you isolated your changes in commits? I don't think such variable will do much.

If the .desktop file is changed, this will cause GNOME to display a blank app icon.

Kimiblock:
I suggest splitting changes into multiple commits, so that testing will be easier.

CoreanXka:
Thank you very much. I will submit a new PR.

PR #4403, opened 2025-08-10
Remove redundant optdepends and bin links to avoid conflicts

CoreanXka:
I reverted to the previous changes and noticed that the icon missing issue no longer appears. Therefore, I gave up on modifying that problem.
Now, I have only removed the `optdepends` content (which originally didn’t exist, and adding it seems to cause a minor cycle dependency).
Deleted the redundant /bin link (I think keeping just one is sufficient for simplicity, and the name “zen” is prone to conflicts),
And removed the SHA256 of one file.

### DIFF
--- a/archlinuxcn/zen-browser-portable/PKGBUILD
+++ b/archlinuxcn/zen-browser-portable/PKGBUILD
@@ -75,4 +75,4 @@ NoDisplay=true''' >"${pkgdir}/usr/share/applications/zen-browser.desktop"
 }
 sha256sums=('e79bba18a1cc0997cab92a1d6cb7d9c8c6c5263de5b1f350b2ab45b1e8ce5d75'
             'b1df66c4246e6584efef74f5c0ae74b7c51c32823bf28bd92a3ef3c938dcaa09'
-            'b016d614f68dd6f51e9b31344bd81befae1de0d2b4f93d9ee4c1ef5c40aff350')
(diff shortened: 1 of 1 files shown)


oldherl:
cycle dependency is not a bug, see phonon-qt6-vlc (provides phonon-qt6-backend) and phonon-qt6

Kimiblock:
The package has no conflict. It requires zen-browser in makedepends to extract package content and conflict it in the actual package. So, the `zen` command link also would not cause any issues. Please build in a chroot environment.

Kimiblock:
Also, because we provide zen-browser, I think the zen command should be present to align with `zen-browser`. Otherwise you will break people's setup with custom .desktop entries or scripts.

In a nutshell, please revert commit [9f41cfe](https://github.com/archlinuxcn/repo/pull/4403/commits/9f41cfe0e8fd6e28f2b0dcd26d6f191755f2e9d1)

Kimiblock:
(from your previous merge request) The StartupWMClass fix could also be included here, it's indeed a bug.

CoreanXka:
The StartupWMClass issue has already been fixed in the last patch submitted to AUR. Currently, two changes have been rolled back, leaving only the SHA256 update. Please decide whether the `optdepends` should be removed. If not, the current state is okay.

Kimiblock:
You can remove the optdepends. I don't use language packs tho.

CoreanXka:
I think it’s fine to keep it as it is, since dependencies can be installed using --adps, so it seems unnecessary to make changes, except for the sha256.

PR #4453, opened 2025-09-17
updpkg: mihomo: make service work more properly

CoelacanthusHex:


archlinuxcn/mihomo/mihomo.service
```diff
@@ -9,10 +9,10 @@ Restart=on-failure
 CapabilityBoundingSet=CAP_NET_ADMIN CAP_NET_RAW CAP_NET_BIND_SERVICE
 AmbientCapabilities=CAP_NET_ADMIN CAP_NET_RAW CAP_NET_BIND_SERVICE
 RestartSec=5
+ExecStartPre=sh -c 'ln -sf "$CONFIGURATION_DIRECTORY/config.yaml" "$STATE_DIRECTORY"/config.yaml && ln -sf "$CONFIGURATION_DIRECTORY/Country.mmdb" "$STATE_DIRECTORY"/Country.mmdb'
+ExecStart=/usr/bin/mihomo -d "$STATE_DIRECTORY"
+ConfigurationDirectory=mihomo::ro
```
Kimiblock, line 14: Is this gonna work? The configuration file is installed as 0700 root and either this will cause pacman to complain when doing -Qkk checks, or the service will not be able to read the configuration file.

I did a `systemd-run -p DynamicUser=yes -P -p ConfigurationDirectory=mihomo::ro -- cat /etc/mihomo/config.yaml` and it reports permission denied.

archlinuxcn/mihomo/mihomo.service
```diff
@@ -9,10 +9,10 @@ Restart=on-failure
...
 StateDirectory=mihomo
-StateDirectoryMode=0700
```
Kimiblock, line 13: Deleting StateDirectoryMode can cause security issues, especially because this contains sensitive proxy server information or subscription links.

Kimiblock (changes requested):

archlinuxcn/mihomo/mihomo.service
CoelacanthusHex: It's a mistake deletion. I will add it back.

Kimiblock:
And there's another issue there which I'm not sure: mihomo seems to skip downloading user-specified GeoIP & GeoSite databases if they are present in the configuration directory. This may cause issues with people who are using third-party rule databases.

CoelacanthusHex:
Ok. The mihomo's strange config-state-data mixed design makes the elegant implementation impossible. But please replace the hardcode `/var/lib/private/mihomo` with `$STATE_DIRECTORY`, please.

Kimiblock:
Moving configurations to StateDirectory makes sense. I'll implement that tomorrow, ping me if I forgot to do it.

PR #4520, opened 2025-11-03
goldendict-ng: add `hunspell-dictionary` as optdepends

DeepChirp:
The official repository provides quite a few Hunspell dictionary packages (for example, [`hunspell-en_us`](https://archlinux.org/packages/extra/any/hunspell-en_us/)), and GoldenDict-NG can use them directly. Perhaps they could be added as optional dependencies?

### DIFF
--- a/archlinuxcn/goldendict-ng/PKGBUILD
+++ b/archlinuxcn/goldendict-ng/PKGBUILD
@@ -27,6 +27,7 @@ depends=(
 	'qt6-speech'
 	'qt6-5compat'
 )
+optdepends=('hunspell-dictionary: Dictionary files for Hunspell support')
 makedepends=('cmake' 'qt6-tools')
 conflicts=('goldendict')
 provides=('goldendict')


archlinuxcn/goldendict-ng/PKGBUILD
```diff
@@ -2,7 +2,7 @@
 
 pkgname=goldendict-ng
 pkgver=25.10.2
-pkgrel=2
+pkgrel=3
```
Integral-Tech:
```suggestion
pkgrel=2
```
The `pkgrel` doesn't need to be updated manually.

DeepChirp: Okay, I might have been influenced by AUR.

DeepChirp:
In addition, the official repository's `libzim` has been updated, and currently, starting the program results in an error message: `symbol lookup error`. I think it should be rebuilt.

Integral-Tech:
Thanks for your contribution :)

PR #4523, opened 2025-11-06
xnec2c: fix pkgver not updated

edward-p:


archlinuxcn/xnec2c/lilac.yaml
CoelacanthusHex: I prefer to use `prefix` option of update_on instead of lstrip manually.

archlinuxcn/xnec2c/lilac.yaml
```diff
@@ -2,7 +2,7 @@ maintainers:
   - github: CoelacanthusHex
     email: coelacanthus@archlinuxcn.org
 build_prefix: extra-x86_64
-pre_build: vcs_update
+pre_build: update_pkgver_and_pkgrel(_G.newver.lstrip('v'))
```
lilydjwg:
```suggestion
pre_build_script: update_pkgver_and_pkgrel(_G.newver.lstrip('v'))
```

PR #4543, opened 2025-11-30
sing-box-alpha: add optdepends

DeepChirp:
`sing-box` can use `sing-geosite` and `sing-geoip` to handle routing rules. Therefore, perhaps it would be better to add them as optional dependencies, just like the `sing-box` package?

### DIFF
--- a/archlinuxcn/sing-box-alpha/PKGBUILD
+++ b/archlinuxcn/sing-box-alpha/PKGBUILD
@@ -13,6 +13,8 @@ url='https://sing-box.sagernet.org/'
 license=('custom:GPL-3.0-or-later WITH name use or association addition')
 
 depends=('glibc')
+optdepends=('sing-geosite-rule-set: GeoSite rule sets'
+            'sing-geoip-rule-set: GeoIP rule sets')
 makedepends=('go' 'clang' 'lld')
 
 source=("${_pkgname}-${_pkgver}.tar.gz::https://github.com/SagerNet/sing-box/archive/v${_pkgver}.tar.gz")


merrkry:
All-in-one geoip/geosite bundles are deprecated by upstream long ago. It is now recommended to use rule_set instead.

DeepChirp:
> All-in-one geoip/geosite bundles are deprecated by upstream long ago. It is now recommended to use rule_set instead.

Yes, you are correct. However, according to the AUR description, taking IP addresses set as an example, both [sing-geoip-rule-set](https://aur.archlinux.org/packages/sing-geoip-rule-set) and [sing-geoip-rule-set-git](https://aur.archlinux.org/packages/sing-geoip-rule-set-git) provide sing-geoip. Therefore, is it reasonable to treat them as optional dependencies?

merrkry:
Pacman's optional depencies are more of a hint. You can still install whatever you want without these lines.

- Old db data is deprecated by upstream and will be deleted and support will be deleted in future versions. Therefore I deleted them (5ddc0153de501b6792a2fbc45927db49ef951ef4) to discourage new users from using it.
- We might want to package rulesets, but the naming is already chaotic. I would prefer let `sing-box` optionally depend on `sing-rule-sets` (something like that, TBD). Point is, it is a completely different format, continuing using the old `geo*` name will mislead users they are the same, or at least compatible. But rulesets require config migration and are not necessarily related to geodbs.

DeepChirp:
In fact, I want to install the ruleset as a dependency, which is more semantic.

The official also provides two data types at the same time, and is still being updated - although db has been replaced. In my opinion, we can actually use sing-rule-sets instead of sing-db, but that would require modifying another software package.

DeepChirp:
@merrkry Due to #4647, the naming of the rule sets seems quite clear. Perhaps `sing-geosite-rule-set` and `sing-geoip-rule-set` could be treated as optional dependencies?

merrkry:
@DeepChirp Sorry for the delay! I've been really busy IRL.
I would suggest you add yourself as maintainer of this package.

PR #4557, opened 2025-12-11
imhex: fix broken build

RocketMaDev:
imhex更新上游又引入了大量变更，动了不少东西

顺便看看能优化哪些参数吧，能用系统的就用系统的了

### DIFF
--- a/archlinuxcn/imhex/PKGBUILD
+++ b/archlinuxcn/imhex/PKGBUILD
@@ -120,9 +120,7 @@ build() {
 
     cmake --build build --parallel
 
-    pushd build
-        ninja unit_tests
-    popd
+    cmake --build build --parallel unit_tests
 }
 
 package() {
@@ -139,8 +137,6 @@ check() {
         python tests/integration/integration.py ../../../../build/imhex --pl
     popd
 
-    pushd build
-        ctest --output-on-failure
-    popd
+    ctest --test-dir build --output-on-failure
 }
 # vim: set ts=4 sw=4 et:


RocketMaDev:
由于现在引入了更多系统依赖，不确定需不需要在依赖更新时rebuild，感觉实在不行的时候手动触发一下也行

archlinuxcn/imhex/PKGBUILD
```diff
@@ -1,62 +1,61 @@
...
-    'libglvnd'
-    'libarchive'
+    'libglvnd' # needed in runtime
     'glibc' 'gcc-libs' 'bash'
-    'libxrandr'
+    'libxrandr' # needed in runtime
     'fontconfig'
+    'capstone'
+    'md4c'
+    # since these "optdepends" are depends of system core components,
```
CoelacanthusHex: If it's optdepends, keep it in optdepends, unless it will cause issue.

archlinuxcn/imhex/PKGBUILD
```diff
@@ -83,46 +76,24 @@ prepare() {
...
-        patch -Np1 -i "$srcdir/patterns-fmt-update.patch"
-    popd
 }
 
 build() {
     cmake -B build -S "$pkgname" \
         -G Ninja \
         -Wno-dev \
-        -D CMAKE_BUILD_TYPE=None \
+        -D CMAKE_BUILD_TYPE=RelWithDebInfo \
```
CoelacanthusHex: Use None to use the flags provoded by Arch Linux.

archlinuxcn/imhex/PKGBUILD
```diff
@@ -131,22 +102,27 @@ build() {
...
         -D USE_SYSTEM_NLOHMANN_JSON=ON \
-        -D USE_SYSTEM_CAPSTONE=OFF \
+        -D USE_SYSTEM_CAPSTONE=ON \
+        -D USE_SYSTEM_MD4C=ON \
+        -D USE_SYSTEM_CLI11=ON \
+        -D USE_SYSTEM_BOOST=ON \
         -D LIBPL_ENABLE_CLI=ON \
         -D LIBPL_BUILD_CLI_AS_EXECUTABLE=OFF \
         -D IMHEX_VERSION="$pkgver" \
+        -D IMHEX_ENABLE_UNIT_TESTS=OFF \
```
CoelacanthusHex: keep tests enabled and do test in check function.

archlinuxcn/imhex/PKGBUILD
RocketMaDev: These are depends in `imhex` or `imhex-bin` from aur, and these dependencies will be loaded via imhex plugins. As plugins are loaded by default, they should be depends.

I could test if dropping them can work, but that requires some hard work.

archlinuxcn/imhex/PKGBUILD
RocketMaDev: Though I got no idea how imhex test. I could check that tomorrow.

archlinuxcn/imhex/PKGBUILD
CoelacanthusHex: You can use `bwrap --ro-bind / / --overlay-src /usr --tmp-overlay /usr bash` to open a shell, then run rm to remove the library file and then test whether imhex can run without it.

RocketMaDev: Only lz4 can be optional. If lz4 not present, imhex automatically disables decompress plugin. The rest must be included or else imhex can not start. Place only lz4 in optdepends? Or keep it as is?

archlinuxcn/imhex/PKGBUILD
```diff
@@ -1,62 +1,61 @@
...
+pkgver=1.38.0
+pkgrel=1
 pkgdesc='A Hex Editor for Reverse Engineers, Programmers and people that value their eye sight when working at 3 AM'
 url='https://imhex.werwolv.net'
 license=('GPL-2.0-only')
 arch=('x86_64')
 depends=(
     'glfw'
-    'hicolor-icon-theme' 'xdg-desktop-portal'
+    'hicolor-icon-theme' 'xdg-desktop-portal' # needed in runtime
```
Xeonacid, line 16: I don't think the comment is needed. Anything in `depends` array should be needed in runtime. :)

archlinuxcn/imhex/PKGBUILD
Xeonacid: > Place only lz4 in optdepends?

I think this approach should be fine.

Xeonacid (commented):
Thanks for the work!

archlinuxcn/imhex/PKGBUILD
RocketMaDev: That's because namcap are giving warnings. I manually checked these deps and confirmed they are used by imhex.

archlinuxcn/imhex/PKGBUILD
```diff
@@ -131,27 +101,46 @@ build() {
...
+        -D IMHEX_ENABLE_UNIT_TESTS=ON \
+        -D IMHEX_ENABLE_PLUGIN_TESTS=ON \
+        -D IMHEX_ENABLE_IMGUI_TEST_ENGINE=ON \
         -D CMAKE_POLICY_VERSION_MINIMUM=3.5
 
     cmake --build build --parallel
+
+    pushd build
+        ninja unit_tests
+    popd
```
CoelacanthusHex:
```suggestion
    cmake --build build --parallel unit_tests
```

archlinuxcn/imhex/PKGBUILD
```diff
@@ -131,27 +101,46 @@ build() {
...
 
+check() {
+    export LD_LIBRARY_PATH="$PWD/build/plugins:$PWD/build/lib/libimhex"
+    pushd "$pkgname/lib/external/pattern_language"
+        python tests/integration/integration.py ../../../../build/imhex --pl
+    popd
+
+    pushd build
+        ctest --output-on-failure
+    popd
```
CoelacanthusHex:
```suggestion
    ctest --test-dir build --output-on-failure
```

PR #4562, opened 2025-12-13
paru: rebuild for libalpm.so.16

q234rty:
This simply updates the alpm.rs crates to https://github.com/archlinux/alpm.rs/commit/b010ec79ee92b54e87f07fa4e287cbabc035c0ca and the pacmanconf crate to 3.1.0 .

This particular commit was chosen because the only remaining commit only bumps the major version of included crates.

Cargo would then still resolve to the old versions of alpm.rs crates for the aur-depends dependency.

AFAIK there's no other way to stop the aur-depends crate from pulling the old versions of alpm.rs without patching aur-depends as well.

I don't know how fast will the paru developer react so I'm leaving this here just in case.


archlinuxcn/paru/PKGBUILD
```diff
@@ -1,23 +1,26 @@
 # Maintainer: Morgan <morganamilo@archlinux.org>
 pkgname=paru
 pkgver=2.1.0
-pkgrel=3
+pkgrel=4
 pkgdesc='Feature packed AUR helper'
 url='https://github.com/morganamilo/paru'
 source=("$pkgname-$pkgver.tar.gz::https://github.com/Morganamilo/paru/archive/v$pkgver.tar.gz"
-        "paru-fix-panic-when-listing-aur.patch")
```
Xeonacid, line 8: Could you please remove the old `paru-fix-panic-when-listing-aur.patch`? Thanks!

q234rty:
A known issue is that during package builds, upgrading an old chroot fails when trying to update pacman since paru binds /etc/makepkg.conf.d as [read only](https://github.com/Morganamilo/paru/pull/1407/files). A workaround would be to remove the chroots under `/var/lib/aurbuild/x86_64` so that paru creates them again, which I guess will work until the next pacman update.

PR #4584, opened 2026-01-01
ttf-plangothic: change LICENSE source

DeepChirp:
The upstream project has changed the font's license file name from `LICENSE.txt` to `LICENSE-OFL.txt`, so the existing PKGBUILD cannot be automatically updated by the bot. Since we only need the font files, not the original code, I don't think it's necessary to include the MIT license.

According to the Arch Linux wiki, it's best not to use `epoch`, so I have removed it.

### DIFF
--- a/archlinuxcn/ttf-plangothic/PKGBUILD
+++ b/archlinuxcn/ttf-plangothic/PKGBUILD
@@ -3,8 +3,8 @@
 pkgbase=ttf-plangothic
 pkgname=(ttf-plangothic otf-plangothic ttc-plangothic otc-plangothic)
 epoch=2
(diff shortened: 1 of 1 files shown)


lilydjwg:
> According to the Arch Linux wiki, it's best not to use `epoch`, so I have removed it.

It can't be removed, because it would cause the package to downgrade.

DeepChirp:
> > According to the Arch Linux wiki, it's best not to use `epoch`, so I have removed it.
> 
> It can't be removed, because it would cause the package to downgrade.

I apologize, I just reread the wiki and realized I misunderstood the meaning of "epoch." I have now added it back.

PR #4589, opened 2026-01-07
Sync paru with x86 version to fix pacman 7 support

yuyichao:
Tested the compilation locally. I left the pkgrel unchanged (and different from the x86 version) to let lilac bump it during rebuild.


Integral-Tech:
@yuyichao  Could you please retain the original coding style of the PKGBUILD?

yuyichao:
Is it intentionally using a different coding style compared to the x86 version? Having two different coding style for the same package in the same repo made it unnecessarily difficult to compare and sync the two version. I don't mind which style it uses but the two should ideally be consistent.

PR #4605, opened 2026-01-16
pytest-runner: rebuild on Python upgrade

Kimiblock:


Kimiblock:
Somethings broken and this is a dupe of #4604

PR #4622, opened 2026-02-03
alarmcn/{v2raya, v2raya-git}: depends on v2raya-core

BenSYZ:
Actually, `v2raya` depends on `v2ray>=5.0.0` or `xray`, use `v2raya-core` to implement such "OR" logic as did in archlinuxcn #3951

### DIFF
--- a/alarmcn/v2raya-git/PKGBUILD
+++ b/alarmcn/v2raya-git/PKGBUILD
@@ -54,7 +54,7 @@ EOF
 }
 
 package() {
-	depends+=('v2ray>=5.0.0')
+	depends+=('v2raya-core')
 
 	cd "${pkgname}/"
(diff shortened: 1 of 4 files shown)


yuyichao:
the core package is already in archlinuxcn and since it has arch ‘any’ it is available for aarch64 as well. The newly added package and the existing version will replace each other in the repo now.

Integral-Tech:
> the core package is already in archlinuxcn and since it has arch ‘any’ it is available for aarch64 as well. The newly added package and the existing version will replace each other in the repo now.

Removed in https://github.com/archlinuxcn/repo/commit/3e4a89470bd984feace3ba514cf825b8cfbd7d42.

yuyichao:
You probably also need to get rid of the repo depends as well. Since the core package is empty and isn’t a build dependency it should just work after deleting the core package from repo depends.

PR #4644, opened 2026-03-02
Add missing dependencies for keypunch-git package

Giteeajake:
Add missing "graphene" and "hicolor-icon-theme" dependencies

### DIFF
--- a/archlinuxcn/telegram-desktop-lily/PKGBUILD
+++ b/archlinuxcn/telegram-desktop-lily/PKGBUILD
@@ -5,7 +5,7 @@
 
 pkgname=telegram-desktop-lily
 _pkgname=telegram-desktop
-pkgver=6.5.1
+pkgver=6.6.0
 pkgrel=1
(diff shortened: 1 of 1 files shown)


archlinuxcn/keypunch-git/PKGBUILD
```diff
@@ -17,7 +17,7 @@ license=('GPL-3.0-or-later')
 provides=('keypunch')
 conflicts=('keypunch')
 depends=('gtk4' 'libadwaita')
-makedepends=('blueprint-compiler' 'cargo' 'git' 'meson' 'ninja')
+makedepends=('blueprint-compiler' 'cargo' 'hicolor-icon-theme' 'git' 'graphene' 'meson' 'ninja')
```
Misaka13514: Maybe they should be added to deps instead of makedeps?

Giteeajake: Let me fix it

Issue #4119, opened 2024-12-21
Orphaning binder_linux-dkms-git


Integral-Tech:
### 问题类型 / Type of issues

* 弃置软件包 / orphaning packages

### 受影响的软件包 / Affected packages

* binder_linux-dkms-git

----
Since `core/linux` & `core/linux-lts` have enabled binderfs support, this package is no longer needed.
Related merge request:
- `core/linux`: https://gitlab.archlinux.org/archlinux/packaging/packages/linux/-/merge_requests/5
- `core/linux-lts`: https://gitlab.archlinux.org/archlinux/packaging/packages/linux-lts/-/merge_requests/1

PR #4666, opened 2026-03-27
alarmcn: remove packages with `any` arch

Integral-Tech:
It is unnecessary to build these packages with `any` arch on the aarch64 build server since they are already built on the x86_64 server.


yuyichao:
It seems that binder_linux-dkms-git was never in the x86 repo? Maybe you want to move it instead?

Integral-Tech:
> It seems that binder_linux-dkms-git was never in the x86 repo? Maybe you want to move it instead?

It was in the x86_64 repo before, but was orphaned and removed in #4119 after binderfs support was added to the kernel packages in Arch Linux official repositories.

PR #4707, opened 2026-04-26
ci: pin dependency version

imlonghao:
1. pin all action to a fixed version
2. upgrade Python to 3.14
3. pin all Python dependency version

Prevent dependency poison

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,12 +9,12 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


lilydjwg:
Then who is going to update them, or they remain forever?

imlonghao:
> Then who is going to update them, or they remain forever?

Our usage is quite simply, I think we can keep it that way for a long time, or unless there is some major update.

lilydjwg:
> there is some major update.

Well, they will come in a few years. I think we can update them when they break.

BTW Copilot only creates noise in the diff view (and makes me find the "Hide comments" toggle).

PR #4737, opened 2026-05-16
dida: add depends

clover-yan:


SilverRainZ:
Most of them are already implicitly satisfied except libappindicator-gtk3, libnotify, libxss, and xdg-utils. Can you please explain why these dependencies are added?

clover-yan:
> Most of them are already implicitly satisfied except libappindicator-gtk3, libnotify, libxss, and xdg-utils. Can you please explain why these dependencies are added?

@SilverRainZ I copied from the DEB package's original dependencies list. AFAIK [`libappindicator-gtk3` is necessary](https://t.me/archlinuxcn_group/3738217). Thinking that I could do it all in one step, I added other packages as well.

PR #4753, opened 2026-05-29
addpkg: nusmv

Yurzi:
nusmv is a new symbolic model checker


Yurzi (commented):
Need to format lilac.yaml

PR #4759, opened 2026-06-04
addpkg: recaf

Yurzi:
Recaf, is a modern Java bytecode editor

I need some reviews and suggestions！

### DIFF
--- a/archlinuxcn/recaf/PKGBUILD
+++ b/archlinuxcn/recaf/PKGBUILD
@@ -1,0 +1,77 @@
+# Maintainer: Yurzi <yurzi@foxmail.com>
+
+pkgname="recaf"
+pkgver=4.0.0
+pkgrel=2
+pkgdesc='A modern Java bytecode editor'
+url='https://github.com/Col-E/Recaf'
+license=('MIT')
+arch=('any')
+
+# Recaf 4.0.0-alpha pins its Jasm dependency to an unreleased JitPack snapshot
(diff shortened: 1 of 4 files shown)


RocketMaDev:
`java-openjfx` does not exist in archlinuxcn, you may consider package it together.

Yurzi:
> `java-openjfx` does not exist in archlinuxcn, you may consider package it together.

Thanks, but is it appropriate to add `java-openjfx` to archlinuxcn?

RocketMaDev:
Adding a package as dependency is acceptable.

Yurzi:
Actually, `recaf` just need jfx to draw UI.
I noticed `hmcl` is also in the archlinuxcn repo. As far as I know, doesn't it also need JFX to run? How does it manage that?
I didn't see any JFX dependency declared in HMCL's related files.

RocketMaDev:
> Actually, `recaf` just need jfx to draw UI. I noticed `hmcl` is also in the archlinuxcn repo. As far as I know, doesn't it also need JFX to run? How does it manage that? I didn't see any JFX dependency declared in HMCL's related files.

It automatically downloads openjfx jar at the first launch.

Yurzi:
I don't think adding `java-openjfx` to archlinuxcn/repo is a good idea, so I'll consider letting `recaf` handle this dependency issue on its own.

Thank you very much for your help.

PR #4760, opened 2026-06-05
addpkg: lwe

YangYuS8:
Add LWE stable package metadata.

Initial version: 0.9.9.

This PR payload was generated from YangYuS8/lwe@38e20ca0730ef630049ac66bda98242e7f8e26cc.

### DIFF
--- a/archlinuxcn/lwe/PKGBUILD
+++ b/archlinuxcn/lwe/PKGBUILD
@@ -1,0 +1,32 @@
+pkgname=lwe
+pkgver=0.9.9
+pkgrel=1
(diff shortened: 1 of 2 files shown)


RocketMaDev:
It's a new package right? Why use "update" instead of "addpkg" in the title?

YangYuS8:
> It's a new package right? Why use "update" instead of "addpkg" in the title?

Yes, this is a new package. Sorry, I used “update” by mistake.

I've renamed the PR title to “addpkg: lwe” and updated the description accordingly. Thanks for pointing it out!

PR #4761, opened 2026-06-05
Update nekobox to 5.11.16

qr243vbi:


RocketMaDev:
Arch Linux already packaged [quickjs-ng](https://gitlab.archlinux.org/archlinux/packaging/packages/quickjs-ng), you can not package a package with same content. You may consider dynamically linking `quickjs-ng`.

qr243vbi:
> Arch Linux already packaged [quickjs-ng](https://gitlab.archlinux.org/archlinux/packaging/packages/quickjs-ng), you can not package a package with same content. You may consider dynamically linking `quickjs-ng`.

Fixed

RocketMaDev:
@qr243vbi `lmdb` is no longer needed? Please remove it from dependencies.

PR #4769, opened 2026-06-10
jellyfin-desktop: rebuild on mpvqt update

Kimiblock:
Jellyfin desktop seems to depend on libMpvQt.so.x with specific version constraint and needs to be rebuilt when said package updates

Context:

jellyfin-desktop: error while loading shared libraries: libMpvQt.so.2: cannot open shared object file: No such file or directory

### DIFF
--- a/archlinuxcn/jellyfin-desktop/lilac.yaml
+++ b/archlinuxcn/jellyfin-desktop/lilac.yaml
@@ -11,3 +11,6 @@ post_build: aur_post_build
 update_on:
   - source: aur
     aur: jellyfin-desktop
+  - source: alpm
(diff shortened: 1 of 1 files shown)


cuihaoleo:
The source code isn't compatible with latest mpvqt: https://build.archlinuxcn.org/imlonghao-api/pkg/jellyfin-desktop/log/1781123721

Since the maintainer is active, I'll wait for a few days to see if they can fix it in a new version: https://aur.archlinux.org/packages/jellyfin-desktop

Kimiblock:
Thanks, maybe depend on a specific version of mpvqt to avoid silent breaking?

PR #4794, opened 2026-06-27
devtools-archlinuxcn: update to 1.5.1

yan12125:
* No rebase conflicts
* [Upstream commits](https://github.com/archlinux/devtools/compare/v1.5.0...v1.5.1) look moderate

### DIFF
--- a/archlinuxcn/devtools-archlinuxcn/PKGBUILD
+++ b/archlinuxcn/devtools-archlinuxcn/PKGBUILD
@@ -5,8 +5,8 @@
 _pkgname=devtools
 pkgname=devtools-archlinuxcn
 epoch=1
-pkgver=1.5.0
-pkgrel=2
+pkgver=1.5.1
+pkgrel=1
 _tag=$pkgver-archlinuxcn1
 _upstream_pkgrel=1
 pkgdesc='Tools for Arch Linux package maintainers, archlinuxcn fork'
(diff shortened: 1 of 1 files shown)


lilydjwg (approved):
It sets machine name now; I hope it won't cause issues.

Not sure what impact not stripping Rust binaries will have, but at least it shouldn't affect non-debugging usage.

yuyichao (approved):

Integral-Tech:
> It sets machine name now; I hope it won't cause issues.
> 
> Not sure what impact not stripping Rust binaries will have, but at least it shouldn't affect non-debugging usage.

Related issue: https://gitlab.archlinux.org/archlinux/packaging/packages/rust/-/work_items/19

Some Rust projects enable stripping of symbols from the binaries in `Cargo.toml`, leading to faulty debuginfo generation. Symbols should be stripped using the `strip` option in the `options` array.

lilydjwg:
> Some Rust projects enable stripping of symbols from the binaries in `Cargo.toml`, leading to faulty debuginfo generation. Symbols should be stripped using the `strip` option in the `options` array.

You mean projects actively make debugging harder for users? That's a pity. And this option can void their attenpt? Then the situation is better than some C++ projects that catches SIGSEGV and not re-raises a signal to have a coredump.

yan12125:
> And this option can void their attenpt?

Seems so. Anyway not stripping binaries should have no negative effects other than larger binaries (if something is not stripped by makepkg).

> It sets machine name now; I hope it won't cause issues.

It shouldn't. Per [systemd-nspawn docs](https://www.freedesktop.org/software/systemd/man/latest/systemd-nspawn.html#-M), the machine name sounds like a field managed by systemd-machind, and should not affect others.

Let's give it a try!

PR #4905, opened 2026-08-29
addpkg: gpu-screen-recorder-nv13.0

Emojigit:
gpu-screen-recorder compiled against FFmpeg using nvcodec 13.0.


Emojigit:
It might be possible to implement a version without self-checking instead of a version bound to ffmpeg 13.0.

archlinuxcn/gpu-screen-recorder-nv13.0/lilac.yaml
DeepChirp, line 1: The package depends on `ffmpeg-nv13.0` from the Arch Linux CN repository, so it is necessary to add `build_prefix` and `repo_depends`.

archlinuxcn/gpu-screen-recorder-nv13.0/REUSE.toml
DeepChirp, line 1: It is best to add a license for the patch as well.

archlinuxcn/gpu-screen-recorder-nv13.0/lilac.yaml
lilydjwg: Not "and" but "or". Prefer to use `repo_depends`.

PR #4916, opened 2026-09-02
fix(waylyrics): OpenCC is no longer used

mokurin000:
Removed opencc from dependencies

### DIFF
--- a/archlinuxcn/waylyrics/PKGBUILD
+++ b/archlinuxcn/waylyrics/PKGBUILD
@@ -2,37 +2,28 @@
 
 pkgname=waylyrics
 pkgver=0.4.2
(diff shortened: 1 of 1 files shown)


archlinuxcn/waylyrics/PKGBUILD
```diff
@@ -2,37 +2,28 @@
 
 pkgname=waylyrics
 pkgver=0.4.2
-_opencc_rust_ver=2.0.0
-pkgrel=1
+pkgrel=2
```
Integral-Tech:
```suggestion
```
Please avoid manually bumping `pkgrel`, since lilac will handle it.

Integral-Tech (approved):

PR #4935, opened 2026-09-15
danxi, danxi-git: depend on org.freedesktop.secrets instead of gnome-keyring

hz-xiaxz:
## What

Replace the hard dependency on `gnome-keyring` with the `org.freedesktop.secrets` virtual package in `danxi` and `danxi-git`, and bump `pkgrel`.

```diff
-depends=('gtk3' 'libsecret' 'gnome-keyring' 'wpewebkit')
+depends=('gtk3' 'libsecret' 'org.freedesktop.secrets' 'wpewebkit')
```

## Why

DanXi stores credentials via `flutter_secure_storage`, whose Linux backend is libsecret. libsecret only needs *some* Secret Service daemon owning `org.freedesktop.secrets` on the session bus; it does not care whether that is gnome-keyring or KWallet.

In Arch both `gnome-keyring` and `kwallet` have `provides=(org.freedesktop.secrets)`, and official packages such as `bitwarden`, `seahorse`, `nheko` and `qtkeychain-qt6` already depend on the virtual package. Hard-depending on `gnome-keyring` forces Plasma users to install a daemon that never runs and cannot be removed.

## Tested

Plasma 6.7 / kwallet 6.30 with `gnome-keyring` installed but not running: `org.freedesktop.secrets` is owned by `ksecretd`, and DanXi's `io.github.danxi_dev.dan_xi/FlutterSecureStorage` item is stored in `kdewallet`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


archlinuxcn/danxi-git/PKGBUILD
```diff
@@ -3,12 +3,12 @@
 pkgname=danxi-git
 _pkgname=${pkgname%-git}
 pkgver=1.5.3.r0.g8bd6792
-pkgrel=1
+pkgrel=2
```
Integral-Tech:
```suggestion
```
Please avoid manually bumping `pkgrel`, since lilac will handle it.

archlinuxcn/danxi/PKGBUILD
```diff
@@ -2,12 +2,12 @@
 
 pkgname=danxi
 pkgver=1.5.3
-pkgrel=1
+pkgrel=2
```
Integral-Tech:
```suggestion
```

hz-xiaxz:
@Integral-Tech Thanks, done

Integral-Tech (approved):

